PR #34, opened 2026-02-10
test: add config factories and expand coverage to 150 tests

aaronteoh:
## Summary
- **TASK-002:** Create reusable factory functions (`make_project_context`, `make_data_source`, `make_dango_config`, etc.) in `tests/factories/` for DRY test data creation. Refactor `tests/conftest.py` to use factories.
- **TEST-001:** Expand config module test coverage from 6 to 150 tests (142 unit + 8 integration) covering all Pydantic models, ConfigLoader, CredentialManager, and exception hierarchy.

## Test breakdown

| File | Tests |
|------|-------|
| `tests/unit/test_factories.py` | 8 |
| `tests/unit/test_config_models.py` | 48 (was 3) |
| `tests/unit/test_config_loader.py` | 35 (was 2) |
| `tests/unit/test_config_credentials.py` | 32 (new) |
| `tests/unit/test_config_exceptions.py` | 5 (new) |
| `tests/integration/test_config_loading.py` | 8 (was 1) |
| **Total** | **150** (was 6) |

## Test plan
- [x] `pytest -v --strict-markers` — 150 passed, 0 failed
- [x] `pytest -m unit` — 142 passed
- [x] `pytest -m integration` — 8 passed
- [x] No source code changes — only test files added/modified

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,1 @@
+# Required for 'from tests.factories...' imports (Python package resolution).
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,44 +3,29 @@
 import pytest
 import yaml
 
-from dango.config.models import (
-    CSVSourceConfig,
-    DangoConfig,
-    DataSource,
-    ProjectContext,
-    SourcesConfig,
-    SourceType,
+from tests.factories.config_factories import (
+    make_dango_config,
+    make_project_context,
+    make_sources_config,
 )
 
 
 @pytest.fixture
 def sample_project_context():
     """A minimal valid ProjectContext instance (no I/O)."""
-    return ProjectContext(
-        name="Test Analytics",
-        created_by="test@example.com",
-        purpose="Unit testing the Dango config system",
-    )
+    return make_project_context()
 
 
 @pytest.fixture
 def sample_sources_config():
     """A SourcesConfig with one CSV source (no I/O)."""
-    return SourcesConfig(
-        sources=[
-            DataSource(
-                name="test_csv",
-                type=SourceType.CSV,
-                csv=CSVSourceConfig(directory="data/test_csv"),
-            )
-        ]
-    )
+    return make_sources_config()
 
 
 @pytest.fixture
 def sample_config(sample_project_context, sample_sources_config):
     """A complete DangoConfig combining project context and sources (no I/O)."""
-    return DangoConfig(
+    return make_dango_config(
         project=sample_project_context,
         sources=sample_sources_config,
     )
--- a/tests/factories/__init__.py
+++ b/tests/factories/__init__.py
@@ -1,0 +1,30 @@
+"""Test factories for creating valid Pydantic model instances.
+
+Usage:
+    from tests.factories.config_factories import make_project_context, make_dango_config
+
+    ctx = make_project_context(name="Custom Name")
+    config = make_dango_config()
+"""
+
+from tests.factories.config_factories import (
+    make_csv_source_config,
+    make_dango_config,
+    make_data_source,
+    make_google_sheets_source_config,
+    make_platform_settings,
+    make_project_context,
+    make_sources_config,
+    make_stakeholder,
+)
+
+__all__ = [
+    "make_csv_source_config",
+    "make_dango_config",
+    "make_data_source",
+    "make_google_sheets_source_config",
+    "make_platform_settings",
+    "make_project_context",
+    "make_sources_config",
+    "make_stakeholder",
+]
--- a/tests/factories/config_factories.py
+++ b/tests/factories/config_factories.py
@@ -1,0 +1,120 @@
+"""Factory functions for config Pydantic models.
+
+Each function returns a valid model instance with sensible defaults.
+All fields are overridable via keyword arguments.
+
+Usage:
+    # Defaults
+    ctx = make_project_context()
+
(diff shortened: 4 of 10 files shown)


aaronteoh:
## Independent Code Review — PR #34

**Reviewer perspective:** Reviewing as an external engineer unfamiliar with the implementation session. Evaluated for correctness, maintainability, test quality, and adherence to project conventions.

---

### Verdict: **Approve** — clean, well-structured work with a few minor observations below.

---

### Strengths

1. **Factory design is clean.** The `{**defaults, **overrides}` pattern is simple, Pythonic, and instantly understandable. No factory library dependency (factory_boy, etc.) — appropriate for this project's scale.

2. **Good separation between `sources` param vs `**overrides` in `make_sources_config`.** The `None`-means-default / `[]`-means-empty convention is documented in the docstring and tested.

3. **Comprehensive model coverage.** All 12+ Pydantic models have dedicated test classes. The parametrized `TestSourceConfigModels` is a pragmatic way to cover 10 rarely-changing models without boilerplate.

4. **Test file organization.** One test file per source module (`test_config_models.py`, `test_config_loader.py`, `test_config_credentials.py`, `test_config_exceptions.py`) maps cleanly to the production code. Easy to navigate.

5. **Markers applied consistently.** `@pytest.mark.unit` on pure logic, `@pytest.mark.integration` on disk I/O classes. No marker misuse.

6. **conftest.py refactoring is backward-compatible.** `sample_project_context`, `sample_sources_config`, `sample_config`, and `tmp_project_dir` all keep the same names and semantics. Any downstream test using these fixtures will work unchanged.

---

### Issues Found

#### Minor (non-blocking)

**1. `tests/__init__.py` — subtle convention tension**

The project convention says "No `__init__.py` in test directories." This PR adds `tests/__init__.py` to enable `from tests.factories...` imports. The plan's rationale is correct (Python's import system needs it for the factories package), but it's worth noting: now that `tests/__init__.py` exists, pytest's test discovery behavior changes slightly — it will use package-based collection. This works fine with `pythonpath = ["."]` in pyproject.toml, but if a future contributor removes `tests/__init__.py` thinking it violates the convention, the factory imports will break. Consider adding a one-line comment in the file explaining why it exists, e.g. `# Required for 'from tests.factories...' imports`.

**2. `make_data_source` type annotation uses `|` union syntax (Python 3.10+)**

```python
def make_dango_config(
    project: ProjectContext | None = None,
    ...
```

pyproject.toml specifies `python_requires = ">=3.10,<3.13"` so this is valid, but worth flagging — `Optional[X]` is used everywhere else in the codebase (models.py, loader.py, credentials.py). The `X | None` style is fine but inconsistent with existing code style.

**3. `TestDeepMerge` instantiates `CredentialManager` just to call `_deep_merge`**

```python
def test_flat_merge(self, tmp_path):
    cm = CredentialManager(tmp_path)
    result = cm._deep_merge({"a": 1}, {"b": 2})
```

This works, but `_deep_merge` is a pure function — the `CredentialManager` instance is only needed because it's an instance method. Not a problem per se, just slightly misleading to future readers (it looks like `tmp_path` matters, but it doesn't). This is the source code's fault, not the test's — flagging for awareness only.

**4. Unused import in `test_config_loading.py`**

```python
from tests.factories.config_factories import make_dango_config, make_data_source
```

`make_data_source` is imported but never used in the integration test file. Should be removed.

**5. Fixture default values changed without documenting the semantic shift**

The `sample_project_context` fixture previously returned `purpose="Unit testing the Dango config system"` and now returns `purpose="Unit testing"`. The commit adapts the one test that asserted the old value, but if any code outside this PR relies on that specific string, it would break silently. Given the project's size and that these are test fixtures, this is very low risk — but good to be aware of.

**6. Some credential tests marked `unit` but do file I/O**

Tests like `TestInitDltDirectory.test_creates_dir_and_files` create actual directories and files via `tmp_path`. Per the pyproject.toml marker definitions: `unit = "no I/O"`, `integration = "may use temp dirs, real files"`. The credential tests use `tmp_path` extensively and arguably belong in integration. However, the loader tests in this same PR (and in the original codebase) also use `tmp_path` with the `unit` marker, so this is a pre-existing convention inconsistency rather than a new one. Not blocking.

---

### Summary

| Category | Count |
|----------|-------|
| Blocking issues | 0 |
| Minor issues | 6 |
| Suggestions | 2 (items 1 and 4 above are easily actionable) |

The PR achieves its goals: factories are well-designed and immediately useful, test coverage jumped from 6 to 150, and no production code was modified. The two actionable items (unused import, missing comment in `tests/__init__.py`) are trivial fixups that could be done in this PR or deferred.

aaronteoh:
## Review Disposition

Fixed in follow-up commit (34caa1b):
- **#1** `tests/__init__.py` — added explanatory comment
- **#4** Unused import — removed `make_data_source` from `test_config_loading.py`

Documented for future follow-up:
- **#4 (marker convention)** — `unit` marker vs `tmp_path` inconsistency tracked in project memory for resolution during TASK-003/004 (CI/pre-commit setup)
- **#6 (source count)** — `SourceType` verified at 33 members; doc inconsistency tracked for next ingestion/ task

No action needed:
- **#2** (`X | None` style) — test-only code, will converge when project-wide style is enforced
- **#3** (`_deep_merge` as instance method) — source code design issue, out of scope for test PR
- **#5** (fixture default change) — no other consumers, already handled

Merging.

PR #60, opened 2026-02-19
Add rate limiting middleware and account lockout (TASK-091)

aaronteoh:
## Summary
- Add pure ASGI rate limiting middleware with per-IP sliding window: login (10/min) and API (200/min) groups, localhost exempt, OPTIONS passthrough
- Add database-backed account lockout functions (`record_failed_login`, `check_account_locked`, `reset_failed_logins`, `unlock_account`) using existing SQLite columns
- Add `AuthConfig` / `RateLimitConfig` / `AccountLockoutConfig` Pydantic models to `config/models.py`
- Wire middleware into `web/app.py` (innermost, after CORS) with graceful config loading fallback
- Re-export lockout functions from `dango.auth`

## New files
- `dango/auth/lockout.py` (172 lines)
- `dango/web/middleware/__init__.py` (8 lines)
- `dango/web/middleware/rate_limit.py` (205 lines)
- `tests/unit/test_auth_lockout.py` (299 lines — 16 tests)
- `tests/unit/test_rate_limit.py` (409 lines — 26 tests)

## Modified files
- `dango/config/models.py` (444→477 lines)
- `dango/web/app.py` (218→236 lines)
- `dango/auth/__init__.py` (89→100 lines)

## Design notes
- Expired lockouts reset the failed attempt counter (users get fresh attempts after waiting out lockout)
- Rate limit middleware is pure ASGI (no `BaseHTTPMiddleware`) to avoid WebSocket/streaming issues
- `_load_rate_limit_config()` catches all exceptions so the app still starts even without a valid project config

## Test plan
- [x] 42 new tests pass (16 lockout + 26 rate limit)
- [x] Full suite: 657 passed, 0 regressions
- [x] ruff lint clean
- [x] ruff format clean
- [x] mypy clean
- [x] All files under 500-line limit

### DIFF
--- a/dango/auth/__init__.py
+++ b/dango/auth/__init__.py
@@ -32,6 +32,12 @@ from dango.auth.database import (
     update_session_activity,
     update_user,
 )
+from dango.auth.lockout import (
+    check_account_locked,
+    record_failed_login,
+    reset_failed_logins,
+    unlock_account,
+)
 from dango.auth.models import APIKey, Role, Session, User, UserCreate, UserResponse, UserUpdate
 from dango.auth.permissions import (
     PERMISSIONS,
@@ -116,6 +122,11 @@ __all__ = [
     "get_api_key_by_hash",
     "list_user_api_keys",
     "update_api_key_last_used",
+    # Lockout functions
+    "check_account_locked",
+    "record_failed_login",
+    "reset_failed_logins",
+    "unlock_account",
     # Permissions
     "PERMISSIONS",
     "ROLE_PERMISSIONS",
--- a/dango/auth/lockout.py
+++ b/dango/auth/lockout.py
@@ -51,6 +51,7 @@ def record_failed_login(
         locked_until_str: str | None = row["locked_until"]
 
         # If already locked and not expired, return remaining time
+        lockout_expired = False
         if locked_until_str is not None:
             locked_until = datetime.fromisoformat(locked_until_str)
             if locked_until.tzinfo is None:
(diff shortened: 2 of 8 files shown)


aaronteoh:
## Code Review — TASK-091 Rate Limiting + Account Lockout

### Findings (all resolved in 47ecab1)

**1. BUG (fixed): Stale `locked_until` prevented re-locking after expiry**
`record_failed_login` reset `current_attempts` in Python when a lockout expired but didn't clear `locked_until` in the DB. Every subsequent call re-detected the stale expired timestamp and re-reset the counter — making it impossible to ever re-lock the account. Fixed with a `lockout_expired` flag that conditionally writes `locked_until = NULL`. Regression test `test_relocks_after_expired_lockout` added.

**2. Medium (fixed): `_maybe_cleanup` was a no-op for inactive IPs**
Cleanup only removed empty deques, but deques from inactive IPs retained stale timestamps that were never pruned (only `_check_rate_limit` prunes, and only for the IP being checked). Fixed to prune expired timestamps from all IPs before checking emptiness. Test rewritten to verify.

**3. Low (fixed): Dead test code in `test_stale_entries_removed`**
First half of the test was a confused attempt that the comments acknowledged didn't work. Replaced with two clean, focused tests.

**4. Low (fixed): `_load_rate_limit_config` silently swallowed exceptions**
Added `logger.debug` so operators can diagnose why rate limiting isn't active.

**5. Low (skipped): `/api/status` prefix match slightly over-broad**
Would also exempt a hypothetical `/api/statusfoo`. No such route exists and none is planned. Not worth the extra complexity of dual exact+prefix matching.

### Verdict

All actionable findings resolved. 659 tests pass, ruff/mypy/format clean, all files under 500 lines. Approved.

PR #106, opened 2026-02-28
P6-003: Add `dango upgrade` command for local version upgrades

aaronteoh:
Title: P6-003: Add dango upgrade command for local version upgrades
Body:
Summary
	∙	New dango upgrade command that upgrades getdango via pip and runs pending database migrations
	∙	--version X.Y.Z flag for specific version installs, --yes flag to skip confirmation prompts
	∙	Pre-flight checks (pip availability, disk space) and backup guidance prompt
	∙	dango status now shows “Update available” notice when a newer version exists on PyPI
	∙	PyPI version check cached 24h in .dango/state/version_check.json; network failures always silent

Test plan
	∙	dango upgrade --help shows correct options
	∙	dango upgrade checks PyPI, shows current vs target version
	∙	--version X.Y.Z installs specific version
	∙	--yes skips confirmation prompts
	∙	Runs pending migrations after pip install
	∙	Clear error message if pip install fails
	∙	Clear error message if migrations fail (with recovery guidance)
	∙	dango status shows “Update available” when newer version exists
	∙	Version check cached for 24h (no repeated PyPI calls)
	∙	Network failures during version check are silent
	∙	Uses sys.executable for pip (not bare pip)
	∙	18 new unit tests pass, full suite (2286 tests) passes
	∙	ruff check + ruff format + mypy clean (no new errors)

### DIFF
--- a/dango/cli/commands/platform.py
+++ b/dango/cli/commands/platform.py
@@ -930,6 +930,25 @@ def status(ctx: click.Context) -> None:
             console.print()
             console.print(f"[dim]Logs: {fastapi_status['log_file']}[/dim]")
 
+        # Check for available updates (cached, non-blocking)
+        try:
+            from dango import __version__ as current_version
+            from dango.cli.commands.upgrade import _get_latest_version_cached
+
+            latest_version = _get_latest_version_cached(project_root)
+            if latest_version:
+                from packaging.version import Version
+
+                if Version(latest_version) > Version(current_version):
+                    console.print()
+                    console.print(
+                        f"[yellow]Update available:[/yellow] "
+                        f"{current_version} → [bold]{latest_version}[/bold]  "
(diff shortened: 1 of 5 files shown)


aaronteoh:
### Prior findings — all resolved:
1. TimeoutExpired not caught on pip install → Fixed
2. Negative cache not written → Fixed (5-min TTL)
3. re unnecessarily lazy-imported → Fixed (module-level)
4. Private helper used cross-module → Renamed to public
5. Status test didn't verify output → Assertions added
6-7. Already addressed in v1
8. dango restore referenced but doesn't exist → Phase 8 comment added
9. No test for pip timeout → test_pip_timeout_shows_error added

Clean implementation, good test coverage (500 lines, 19 test cases), follows project patterns."

PR #132, opened 2026-03-12
P5-001: Add PostgreSQL to source registry

aaronteoh:
## Summary
- Add PostgreSQL as the 34th source type with full wizard support
- Uses connection URL pattern (matching MongoDB) — zero changes needed to dlt_runner.py, config/models.py, or source_wizard.py
- Includes schema filtering, table selection, setup guide with read-only user creation, and pip dependency declarations (sqlalchemy + psycopg2-binary)
- Update source count from 33 → 34 across all docs (CLAUDE.md, ARCHITECTURE.md, ingestion/CLAUDE.md, config/models.py)

## Test plan
- [ ] Pre-commit hooks pass (ruff check, ruff format, mypy, file headers, docstrings)
- [ ] `dango source add` → PostgreSQL appears under Databases category
- [ ] Complete wizard → verify `.dango/sources.yml` has postgres entry with `credentials_env`
- [ ] Optional: sync against a real PostgreSQL instance

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -4,7 +4,7 @@
 
 Dango is an open-source data platform for small teams that integrates four production-grade tools into a single `pip install` + `dango start` workflow:
 
-- **dlt** (data load tool) for ingestion from 33 source types
+- **dlt** (data load tool) for ingestion from 34 source types
 - **DuckDB** as the embedded analytical warehouse
 - **dbt** for SQL-based data transformation
 - **Metabase** for dashboards and visualization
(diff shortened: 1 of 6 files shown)


aaronteoh:
## Code Review — Approved

### Round 1

**Registry entry (registry.py):** Correct. `credentials_env` → `credentials` matches dlt's `sql_database()` signature (verified via `inspect.signature`). `schema` and `table_names` match exact param names. Built-in import path (`dlt.sources.sql_database`) works via existing fallback at dlt_runner.py:1149-1153. `pip_dependencies` correctly declares sqlalchemy + psycopg2-binary.

**3 issues found and fixed:**

1. **generic_config fallback (blocking):** Wizard wrote `config["postgres"] = params` but `DataSource` has no `postgres` field — Pydantic V2 silently drops it. Fixed: check `DataSource.model_fields` dynamically, fall back to `generic_config` for sources without dedicated fields.

2. **List param parsing (pre-existing, newly exposed):** No handler for `type: "list"` in wizard — `table_names` would be passed as string instead of `List[str]`. Fixed: added `"list"` type handler that splits comma-separated input.

3. **Category comment (cosmetic):** Restored explanatory comment about postgres using built-in sql_database.

### Round 2

Verified fixes are correct. Caught ordering issue: list parsing ran before skip/empty guards, meaning whitespace-only input would produce `[]` (no tables) instead of `None` (all tables). Fixed: moved list parsing after guards, added `or None` fallback.

### Round 3

Final diff reviewed. All edge cases handled:
- `""` → `None` (all tables) ✓
- `"  "` → `None` (all tables) ✓  
- `"table1, table2"` → `["table1", "table2"]` ✓
- `generic_config` used for non-dedicated sources ✓
- Existing wizard-enabled sources unaffected ✓

**No remaining tech debt.** LGTM.

PR #134, opened 2026-03-12
P5-010: Parameterize Mux views date range and fix idempotency

aaronteoh:
## Summary
- Add `start_date`/`end_date` parameters to `mux_source()` and `views_resource()` (default: 30 days ago → today; was hardcoded to yesterday only)
- Switch `views_resource` write disposition from `"append"` to `"replace"` for idempotent re-runs
- Add `start_date` optional param to Mux registry entry in `registry.py`

## Test plan
- [x] `grep "subtract(days=1)"` confirms hardcoded date removed
- [x] `grep "write_disposition"` confirms `"replace"` for views_resource
- [x] All 2813 tests pass (`pytest`)
- [x] Pre-commit hooks pass

### DIFF
--- a/dango/ingestion/dlt_sources/mux/__init__.py
+++ b/dango/ingestion/dlt_sources/mux/__init__.py
@@ -13,15 +13,22 @@ from .settings import API_BASE_URL, DEFAULT_LIMIT
 
 
 @dlt.source
-def mux_source() -> Iterable[DltResource]:
+def mux_source(
+    start_date: str | None = None,
+    end_date: str | None = None,
+) -> Iterable[DltResource]:
     """
-    Source function that passes all video assets and every video view from yesterday to be loaded.
+    Source function that loads all video assets and video views for a configurable date range.
+
+    Args:
+        start_date: Start of the date range (ISO 8601 string or datetime). Defaults to 30 days ago.
+        end_date: End of the date range (ISO 8601 string or datetime). Defaults to today.
 
     Yields:
         DltResource: Video assets and video views to be loaded.
     """
     yield assets_resource
-    yield views_resource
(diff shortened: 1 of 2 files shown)


aaronteoh:
## Code Review: Approve

**Verdict: Ship it.**

### What's correct

1. **Core fix is sound.** Hardcoded yesterday-only range replaced with configurable 30-day default. `"append"` → `"replace"` eliminates duplicates. Both are the right calls.

2. **Return type annotation fix.** `views_resource` previously declared `-> Iterable[DltResource]` but actually yields `TDataItem`. The PR corrects this to `-> Iterable[TDataItem]`.

3. **Date parsing is defensive.** `pendulum.parse(str(x))` handles both `str` and `datetime` inputs correctly. The `str()` wrapper is necessary because `dlt_runner._build_source_config()` (line 993-996) passes stdlib `datetime` objects, not strings.

4. **`assets_resource` untouched.** Correct — asset metadata isn't time-series; `write_disposition="merge"` is appropriate.

5. **Registry change is minimal and correct.** Only `start_date` exposed (not `end_date`). Defaulting end to "today" is the right UX.

### One minor note (non-blocking)

Type annotation says `str | None` but actual runtime input from `dlt_runner` is `datetime`. The `isinstance(x, pendulum.DateTime)` branch is dead code in practice — but harmless as defensive coding, and the `pendulum.parse(str(x))` path handles all actual inputs correctly.

### Pre-existing issues (not introduced by this PR, out of scope)

- Double `response.json()` call per loop iteration (lines 85-87)
- No pagination guard

All 2813 tests pass. Minimal, focused change that fixes a real data-loss bug and an idempotency issue.

PR #164, opened 2026-04-10
P8 Round 3: Fix Metabase DuckDB connection (read_only + :ro mount)

aaronteoh:
## Summary

- Set `read_only: True` in Metabase DuckDB connection config (`metabase.py:725`)
- Add `:ro` suffix to data volume mount in docker-compose template

**Root cause:** DuckDB file locking (`fcntl`/`flock`) does not work reliably over Docker Desktop's bind mount FUSE layer on macOS. With `read_only: False`, the DuckDB driver hangs trying to acquire a write lock until timeout (10s), causing `dango start` to fail with "Cannot connect Metabase to DuckDB".

**Why read_only is correct:** Metabase is a BI tool that only reads data. All writes go through dlt (sync) and dbt (transform) on the host. The `:ro` mount is defense-in-depth.

## Test plan

- [x] `pytest` — 3017 passed
- [ ] Manual: `dango init test-project && cd test-project && dango start` → Metabase connects without timeout → web UI loads

### DIFF
--- a/dango/templates/docker-compose.yml.j2
+++ b/dango/templates/docker-compose.yml.j2
@@ -10,7 +10,7 @@ services:
       - "{{ metabase_port }}:3000"
     volumes:
       - metabase-data:/metabase-data
-      - ./data:/data
+      - ./data:/data:ro
       - ./metabase-plugins:/app/plugins
     environment:
       MB_DB_TYPE: h2
--- a/dango/visualization/metabase.py
+++ b/dango/visualization/metabase.py
@@ -722,7 +722,7 @@ def setup_metabase(
             "details": {
                 "database_file": docker_duckdb_path,
                 "old_implicit_casting": True,
-                "read_only": False,
+                "read_only": True,
                 # Note: DuckDB driver doesn't support schema-filters-type/patterns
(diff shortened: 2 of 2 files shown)


aaronteoh:
## Code Review — Approve ✓

**Diff correctness:**
- `metabase.py:725`: `read_only: False` → `True` — correct. The only place the initial DuckDB connection config is created for Metabase. The companion code in `auth/metabase_sync.py:415` already sets `read_only = True` (for user sync flows), so this aligns them.
- `docker-compose.yml.j2:13`: `./data:/data` → `./data:/data:ro` — correct. Defense-in-depth. No other templates reference this mount.

**Completeness check:**
- No other `read_only: False` references remain in `visualization/`
- No other `./data:/data` mounts without `:ro` exist in the codebase
- `cli/init.py` renders from the Jinja2 template (`_create_docker_compose` uses `PackageLoader`), so the template fix covers all generated docker-compose files
- No behavioral side-effects: Metabase only reads DuckDB data; writes go through dlt/dbt on the host

**Risk:** Minimal. Two one-line changes, both making Metabase more restrictive (read-only).

PR #167, opened 2026-04-13
fix: Metabase SSO auto-bridging, CSRF exemption, 403 handling (BUG-011/014/015)

aaronteoh:
## Summary

- **BUG-011 (SSO bridging):** Auto-bridge Metabase session when user has no `metabase.SESSION` cookie, instead of waiting for a 401 that never comes (Metabase returns login page as 200). Lazy-sync users to Metabase on first proxy access if startup sync failed.
- **BUG-014 (proxy CSRF):** Exempt `/metabase/` proxy routes from Dango CSRF check (Metabase's own frontend sends POSTs without Dango headers). Fix `_build_response()` to preserve multiple `Set-Cookie` headers from Metabase.
- **BUG-015 (permission denied):** Re-bridge on 403 (not just 401) so stale sessions from direct Metabase access at `:3000` are transparently refreshed.

### Files changed
- `dango/auth/metabase_bridge.py` — lazy sync fallback in `bridge_metabase_login()`
- `dango/web/routes/metabase_proxy.py` — auto-bridge on missing session, 403 handling, multi-header fix
- `dango/web/middleware/auth.py` — CSRF exemption for `/metabase/` prefix
- 3 test files — 9 new tests (3 bridge, 5 proxy, 1 middleware)

### Security note
CSRF middleware change is narrowly scoped: only `/metabase/` prefix exempt. Dango auth still required. Metabase has its own CSRF protection. `metabase.SESSION` cookie is HttpOnly + SameSite=Lax.

## Test plan
- [x] 58 targeted tests pass (bridge + proxy + middleware)
- [x] 551 full auth test suite passes — no regressions
- [x] ruff lint + format clean
- [x] mypy clean (all 3 source files)
- [ ] Manual: `dango start` → login → click Metabase link → auto-login (no Metabase login page)
- [ ] Manual: Verify `metabase.SESSION` cookie set in browser after first `/metabase/` access

### DIFF
--- a/dango/auth/metabase_bridge.py
+++ b/dango/auth/metabase_bridge.py
@@ -56,21 +56,56 @@ async def bridge_metabase_login(
     user: User,
     project_root: Path,
     metabase_url: str | None = None,
+    *,
+    db_path: Path | None = None,
 ) -> str | None:
     """Create a Metabase session for *user* using their encrypted password.
 
     Returns the Metabase session ID on success, or ``None`` if bridging
     cannot proceed (no credentials, Metabase down, etc.).
+
+    When *db_path* is provided and the user has no Metabase credentials,
+    attempts a lazy sync (creating the Metabase account on the fly) before
+    giving up.  This handles the case where Metabase was not ready at
+    startup and the initial sync silently failed.
     """
     try:
         if user.metabase_password_enc is None or user.metabase_user_id is None:
-            logger.debug("Metabase bridge skip: no credentials for %s", user.email)
-            return None
+            if db_path is None:
+                logger.debug("Metabase bridge skip: no credentials for %s", user.email)
+                return None
+
+            # Lazy sync: user was never synced to Metabase
+            logger.info("Metabase bridge: attempting lazy sync for %s", user.email)
+            if metabase_url is None:
+                metabase_url = await get_metabase_url(project_root)
+            if metabase_url is None:
+                return None
+
+            from dango.auth.metabase_sync import sync_user_to_metabase
+
+            mb_id = await asyncio.to_thread(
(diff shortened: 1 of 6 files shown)


aaronteoh:
## Code Review

Reviewed all 6 changed files independently. Two commits: initial implementation + review fixes.

### Findings addressed (review fix commit)
- **Stale cookie after failed re-bridge** — When 401/403 re-bridge fails, the stale `metabase.SESSION` cookie is now cleared so the next request uses the auto-bridge path (which enables lazy sync) instead of looping on the same 401/403.
- **Duplicate cookie-setting code** — Extracted `_set_mb_session_cookie()` and `_clear_mb_session_cookie()` helpers.
- **Missing test coverage** — Added `test_lazy_sync_no_metabase_url` (lazy sync without metabase.yml) and `test_non_metabase_path_still_requires_csrf` (CSRF exemption path boundary).

### Findings assessed as not actionable
- **Lazy sync race condition** — Two concurrent requests for an unsynced user could both trigger `sync_user_to_metabase()`. Accepted: the sync function checks `metabase_user_id` before creating, Metabase has a unique email constraint, and this only happens once per user lifetime.
- **CSRF exemption scope** — `/metabase/` prefix exemption is safe because `dango_session` cookie is SameSite=Lax (browsers block cross-site POST). Dango auth is still required.
- **No `asyncio.wait_for` on lazy sync** — Each internal HTTP call in `sync_user_to_metabase` has its own 10s timeout. Total bounded at ~30-40s worst case.

### Verification
- 553 auth tests pass (12 new), no regressions
- ruff lint + format clean
- mypy clean on all 3 source files
- All pre-commit hooks pass

LGTM ✅

PR #176, opened 2026-04-18
R7-A: Systematic null guard fix for app.js, logs.js, catalog.html (BUG-053)

aaronteoh:
## Summary

- Fix 80 unguarded `getElementById`/`querySelector` calls across `app.js` (54 violations) and `logs.js` (26 violations) that crash on pages missing the expected DOM elements
- Fix `catalog.html` Alpine.js null-reference issue by wrapping the detail panel in `<template x-if="selectedModel">` instead of `x-show` (prevents expression evaluation when `selectedModel` is null)
- Add `showToast()` container auto-creation for pages that don't have a `#toast-container` in their template
- Add `("openMetabase", "loginForm")` to the pre-commit checker allowlist (false positive — `loginForm` is inside an HTML template literal written to a new browser window)

## Test plan

- [x] `scripts/check_js_null_guards.py` exits 0 (was 80 violations)
- [x] `pytest` — 3063 passed, 30 skipped
- [x] All pre-commit hooks pass (including JS null guard check)
- [ ] Manual: load Sources, Models, Catalog, Logs pages — no console errors
- [ ] Manual: catalog detail panel shows/hides without Alpine.js null errors

### DIFF
--- a/dango/web/static/js/app.js
+++ b/dango/web/static/js/app.js
@@ -1424,7 +1424,7 @@ function showToast(message, type = 'info') {
     if (!container) {
         container = document.createElement('div');
         container.id = 'toast-container';
-        container.className = 'fixed top-4 right-4 z-50 space-y-2';
+        container.className = 'fixed bottom-4 right-4 space-y-2 z-50';
         document.body.appendChild(container);
     }
 
@@ -1955,9 +1955,9 @@ async function openSourceDetail(sourceName) {
 
         // Display configuration with custom rendering for different source types
         const configElement = document.getElementById('detail-config');
-        if (configElement) {
-
-        if (details.config.type === 'stripe' && details.config.stripe) {
+        if (!configElement) {
+            // Config element not on this page
+        } else if (details.config.type === 'stripe' && details.config.stripe) {
             // Clean rendering for Stripe sources - only show user-relevant info
             const stripeConfig = details.config.stripe;
             const endpoints = stripeConfig.endpoints || [];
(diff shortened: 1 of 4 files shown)


aaronteoh:
## Code Review — Approved

Reviewed the full diff (2 commits, 4 files, +171/−77 lines).

### What was checked

- All 80 violations from `check_js_null_guards.py` are resolved (54 in app.js, 26 in logs.js)
- Guard patterns are appropriate: early returns for page-gating elements, optional chaining for classList toggles, individual guards for property assignments
- `catalog.html` correctly uses `<template x-if="selectedModel">` instead of `x-show` to prevent Alpine.js null-reference evaluation
- `showToast()` container-creation fallback uses `bottom-4 right-4 space-y-2 z-50` matching all 4 template definitions (dashboard, sources, models, logs)
- Allowlist entry for `openMetabase`/`loginForm` false positive is correct — element is inside an HTML template literal written to a new browser window
- No behavioral changes on pages where elements exist (guards are no-ops)
- `check_js_null_guards.py` exits 0
- 3063 tests pass, 30 skipped
- All pre-commit hooks pass

### Issues found and fixed in second commit

1. **Bug: toast container position** — fallback used `top-4` but all templates use `bottom-4`. Fixed.
2. **Structure: configElement guard** — `if (configElement) { ... } // end` with mismatched indentation. Restructured as chained `if (!configElement) {} else if (...)`.
3. **Gap: logs.js showToast()** — same unguarded pattern (13 lines from assignment, outside checker's 10-line window). Added same container-creation fallback.
4. **Nit: catalog.html indentation** — `<template>` tags aligned with parent `<div>`.

### Verdict

All issues resolved. No remaining concerns. Approved for merge.

PR #179, opened 2026-04-19
fix: use correct Metabase API endpoint for password updates (BUG-011)

aaronteoh:
## Summary

- **BUG-011 (Critical):** Fixed Metabase SSO session bridging broken by wrong password update endpoint. `PUT /api/user/{id}` silently ignores the `password` field — must use `PUT /api/user/{id}/password`. Fixed in both code paths: `_link_metabase_admin()` in `platform/common/startup.py` and the fallback path in `sync_user_to_metabase()` in `auth/metabase_sync.py`.
- **BUG-021:** No changes needed — all `dango.local` URL references were already removed in prior fix rounds. The two remaining `admin@dango.local` references are intentional exclusion filters (per plan).

## Files changed

- `dango/auth/metabase_sync.py` — line 331: `/api/user/{id}` → `/api/user/{id}/password`
- `dango/platform/common/startup.py` — line 336: `/api/user/{id}` → `/api/user/{id}/password`

## Test plan

- [x] `pytest tests/unit/test_platform_startup.py tests/unit/test_metabase_sync.py tests/unit/test_metabase_bridge.py` — 76 passed
- [x] `grep -n "api/user.*password" dango/auth/metabase_sync.py dango/platform/common/startup.py` — both include `/password` suffix
- [x] `grep -rn "dango.local" dango/` — only filter/exclusion contexts remain

### DIFF
--- a/dango/auth/__init__.py
+++ b/dango/auth/__init__.py
@@ -64,6 +64,7 @@ from dango.auth.metabase_sync import (
     sync_all_users_to_metabase,
     sync_user_role,
     sync_user_to_metabase,
+    update_metabase_user_password,
 )
 from dango.auth.models import APIKey, Role, Session, User, UserCreate, UserResponse, UserUpdate
 from dango.auth.oauth_login import (
@@ -199,6 +200,7 @@ __all__ = [
     "sync_all_users_to_metabase",
     "sync_user_role",
     "sync_user_to_metabase",
+    "update_metabase_user_password",
     # Permissions
     "PERMISSIONS",
     "ROLE_PERMISSIONS",
--- a/dango/auth/metabase_sync.py
+++ b/dango/auth/metabase_sync.py
@@ -328,7 +328,12 @@ def sync_user_to_metabase(
             # Update password on existing Metabase user so we can bridge SSO
             password = generate_metabase_password()
             if (
-                _mb_put(metabase_url, session, f"/api/user/{mb_user_id}", {"password": password})
+                _mb_put(
+                    metabase_url,
+                    session,
+                    f"/api/user/{mb_user_id}/password",
+                    {"password": password},
+                )
                 is None
             ):
                 logger.warning("Failed to update Metabase password for user %s", mb_user_id)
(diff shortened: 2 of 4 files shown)


aaronteoh:
## Independent Code Review — 3 rounds

**Reviewer:** Claude (independent review role)
**Result: Approved ✓**

---

### What the PR does

Fixes BUG-011 (Critical): Metabase SSO session bridging was permanently broken because both `_link_metabase_admin()` and `sync_user_to_metabase()` called `PUT /api/user/{id}` with `{"password": ...}`. Metabase silently accepts this request and returns 200 but ignores the password field entirely. The correct endpoint is `PUT /api/user/{id}/password`.

BUG-021: Confirmed already resolved in prior rounds — two remaining `admin@dango.local` references are intentional exclusion filters, no changes needed.

---

### Round 1 findings → all resolved

- **HIGH:** `startup.py` was missing `old_password` in the password update payload. Metabase requires `old_password` when the session user is the same as the target user (admin self-change). Fixed by extracting `update_metabase_user_password()` and passing `old_password=mb_password`.
- **MEDIUM:** `startup.py` used raw `requests.put()` while `metabase_sync.py` used `_mb_put()`. Fixed by extracting a shared public helper `update_metabase_user_password()` — both callers now use it, eliminating the duplication.
- **LOW:** Tests mocked `requests.put` without asserting on the URL path or payload. Fixed — `test_happy_path_links_admin` now asserts `/api/user/42/password` in the URL and `old_password == "testpw"` in the payload.

### Round 2 findings → resolved + one intentional deferral

- **LOW:** `status=pw_resp.status_code` was dropped from the failure log when the raw `requests.put()` call was replaced. Fixed by adding a `warning` log inside `update_metabase_user_password()` when `_mb_put` returns `None`, covering silent non-200 responses.
- **LOW/observation:** `sync_user_to_metabase()` fallback doesn't pass `old_password`. Correct for its primary use case (admin changing another user's password). The edge case where the admin user themselves hits this path is pre-existing, handled by `_link_metabase_admin()` as the designated code path, and not worth the complexity of fixing in this PR.

### Round 3

No new issues found. The Round 2 logging fix is a 4-line addition to a function reviewed twice — clean.

---

**Final state:** 3 commits, 76 tests passing, all pre-commit hooks green.

PR #180, opened 2026-04-19
R7-D: Fix 2FA QR code, recovery text, and API key revoke

aaronteoh:
## Summary

- **BUG-059**: Add QR code rendering in 2FA setup — adds `qrcodejs` via CDN, renders `provisioning_uri` as a scannable QR code in the 'qr' step. Manual secret kept as 'Can't scan?' collapsible fallback.
- **BUG-060**: Fix recovery codes message during first-time 2FA setup. Root cause: `verify2FASetup()` set `totp_enabled = true` before transitioning to 'recovery' step, causing Alpine to render the regeneration block ("Your old recovery codes are no longer valid") during first setup. Fixed by adding `twoFaIsFirstSetup` flag — text is now conditional on setup vs regeneration path.
- **BUG-061**: Add `confirm()` dialog before API key revocation to prevent accidental deletion.

## Files changed
- `dango/web/templates/account.html` (only file)

## Test plan
- [x] All 2961 unit tests pass (`pytest tests/unit/`)
- [ ] Manual: Enable 2FA → QR code renders, "Can't scan?" shows secret as fallback
- [ ] Manual: Verify 2FA → recovery codes say "Save your recovery codes" (not "old codes")
- [ ] Manual: Regenerate recovery codes → recovery codes say "New recovery codes" / "old codes" message
- [ ] Manual: Revoke API key → confirm dialog appears before deletion

### DIFF
--- a/dango/web/templates/account.html
+++ b/dango/web/templates/account.html
@@ -222,26 +222,6 @@
                         </form>
                     </div>
 
-                    <!-- Step: recovery codes -->
-                    <div x-show="twoFaStep === 'recovery'">
-                        <div class="p-3 bg-yellow-50 border border-yellow-200 rounded-md mb-3">
-                            <p class="text-sm font-medium text-yellow-800 mb-1">Save your recovery codes</p>
-                            <p class="text-xs text-yellow-700">Store these codes safely. Each can be used once if you lose access to your authenticator.</p>
-                        </div>
-                        <div class="mb-3 p-3 bg-gray-50 rounded-md border font-mono text-sm grid grid-cols-2 gap-1">
-                            <template x-for="code in twoFaRecoveryCodes" :key="code">
-                                <span x-text="code"></span>
-                            </template>
-                        </div>
-                        <div class="flex space-x-2">
-                            <button @click="copyRecoveryCodes()"
-                                    class="px-3 py-2 text-sm bg-gray-100 rounded-md hover:bg-gray-200">Copy</button>
-                            <button @click="downloadRecoveryCodes()"
-                                    class="px-3 py-2 text-sm bg-gray-100 rounded-md hover:bg-gray-200">Download</button>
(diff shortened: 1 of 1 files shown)


aaronteoh:
## Independent Code Review — R7-D

**Reviewed:** 2 commits (`f40c1a7`, `c3ca67b`)

### BUG-059 (QR Code)
- ✅ `provisioning_uri` was already stored in `twoFaUri` — QR rendering correctly uses `$nextTick()` after step transition so the container exists in DOM
- ✅ SRI hash added (`sha256-xUHvBjJ4hahBW8qN9gceFBibSFUzbe9PNttUvehITzY=`) with `crossorigin="anonymous"` — supply-chain risk on a 2FA page addressed
- ✅ `typeof QRCode !== 'undefined'` guard handles CDN load failure gracefully; "Can't scan?" fallback remains accessible
- ✅ No CSP configured in Caddy templates or FastAPI middleware — no compatibility issue

### BUG-060 (Recovery Text)
- ✅ Root cause correctly identified: `verify2FASetup()` set `totp_enabled=true` before step transition, causing Alpine to render the regeneration block ("old codes") during first setup
- ✅ `twoFaIsFirstSetup` flag is set before `totp_enabled=true` assignment — ordering is correct and documented with a comment
- ✅ Dead recovery block inside `x-if="!user.totp_enabled"` removed — it was unreachable since `totp_enabled` becomes `true` before the `'recovery'` step is ever reached via `verify2FASetup()`
- ✅ `reset2FA()` clears the flag; regeneration path correctly leaves flag as `false`

### BUG-061 (Revoke Confirmation)
- ✅ `confirm()` placed before the API call, not just before the UI update
- ✅ `revokeAllOtherSessions()` also given a matching confirmation for consistency (the more destructive bulk operation was previously unguarded)

### Verdict
All issues resolved. No new concerns introduced. **Approved.**

PR #193, opened 2026-04-27
R8-F: Fix deploy pipeline integration bugs

aaronteoh:
## Summary

- **BUG-098:** Sync `.dango/project.yml` to server (was missing from `SYNC_CONFIG_FILES`)
- **BUG-100:** Fix admin email not reaching Metabase setup under systemd — `load_dotenv()` in `serve.py`, persist `DANGO_ADMIN_EMAIL` to remote `.env`, filter `admin@localhost` and dot-less domains in `startup.py`
- **BUG-110:** Document intentional `metabase-plugins/` exclusion from sync
- **BUG-114:** Generate `dbt/profiles.yml` during cloud deploy (DO uses size tier specs, BYOS auto-detects hardware via SSH)
- **Docker rebuild:** `push_deploy()` detects Docker file changes in `synced_files` and uses `docker compose up --build -d` instead of `start`
- **BUG-111/113:** Verified already fixed by R8-E/R8-B respectively

## Test plan

- [x] 111 targeted tests pass across 4 test files (15 new tests added)
- [x] Full unit suite passes (all background runs exit 0)
- [x] All pre-commit hooks pass (ruff, ruff-format, mypy, file-size-check)
- [x] File-exemptions.yml and CLAUDE.md line counts updated
- [ ] Cloud deploy retest in Session 8 (manual, post-merge)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,7 +91,7 @@ dango/                          # Python package source
 │   │   ├── serve.py            # serve production foreground server
 │   │   ├── deploy.py           # deploy group (wizard default, --byos, destroy)
 │   │   ├── deploy_wizard.py    # Interactive deploy wizard + BYOS (808 lines)
-│   │   ├── deploy_provision.py # Provisioning orchestration + BYOS (737 lines)
+│   │   ├── deploy_provision.py # Provisioning orchestration + BYOS (813 lines)
 │   │   ├── migrate.py          # migrate group (status, run)
 │   │   ├── remote.py           # remote group + push/rollback/firewall/domain (698 lines)
 │   │   ├── remote_env.py       # remote env subgroup (set/get/list/delete)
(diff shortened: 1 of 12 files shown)


aaronteoh:
## Code Review — Approve ✓

Reviewed all 12 changed files (5 source, 4 tests, 3 docs). No blocking issues found.

### Per-file assessment

| File | Verdict | Notes |
|------|---------|-------|
| `file_sync.py` | ✓ | `project.yml` correctly placed in `.dango/` group |
| `serve.py` | ✓ | `load_dotenv()` silently handles missing `.env` (returns False). Default `override=False` is correct — explicit systemd env vars take precedence |
| `startup.py` | ✓ | `admin@localhost` filter clean. Domain dot-check catches exact Metabase failure mode. Minor style nit: `_get_logger2` aliasing is slightly awkward but functional (mutually exclusive branches) |
| `deploy_provision.py` | ✓ | `grep -q \|\| echo >>` correctly handles all 3 cases: var exists (no-op), file exists without var (appends), file missing (bash `>>` creates it). Email validated by `_EMAIL_RE` before shell interpolation — no injection risk. `_generate_cloud_profiles` reuses `resize.py`'s `generate_dbt_profiles_yml` — no duplication |
| `deployer.py` | ✓ | `_DOCKER_FILES` matches SYNC_CONFIG_FILES entries. `rebuild_docker` is keyword-only with `False` default. Detection via `set(synced_files) & _DOCKER_FILES` works because synced_files uses local-relative paths. Edge case: default empty SyncResult → no rebuild. Correct |
| Tests (4 files) | ✓ | 15 new tests, 111 total pass. Fix to `test_returns_errors_when_setup_fails_before_duckdb` (`bad@localhost` → `bad@example.com`) is correct — old email now caught by new domain validation |
| Docs (3 files) | ✓ | Line counts verified via `wc -l` |

### Pre-existing note (not introduced by this PR)

Non-rebuild path in `_start_all_services` runs `docker compose start metabase` as root, while rebuild path correctly uses `sudo -u dango`. Harmless (root can run Docker commands), but could be cleaned up in a future PR.

PR #195, opened 2026-04-30
R9-B: Fix Metabase password desync and container name mismatch

aaronteoh:
## Summary

- **BUG-115:** `_link_metabase_admin()` now writes the new Metabase admin password back to `metabase.yml` after updating it in Metabase and auth.db, preventing subsequent API calls from using a stale password and getting 401 Unauthorized.
- **BUG-118:** `refresh_metabase_connection()` now uses `DockerManager.compose_project_name` (hash-based naming) instead of the old `{dirname}-metabase-1` pattern, so the container is actually found.

## Test plan

- [x] 4 new tests: metabase.yml write-back on success, no write-back on failure, hash-based container name, container-not-found path
- [x] All 3075 unit tests pass
- [x] All pre-commit hooks pass (ruff, mypy, file-size-check, etc.)
- [x] File exemptions updated (`metabase.py` 1152→1151, `test_platform_startup.py` added at 601)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -334,7 +334,7 @@ Full exemption registry: [`docs/file-exemptions.yml`](docs/file-exemptions.yml)
 | `ingestion/dlt_runner.py` | 2373 | — (exempt, too risky) |
 | `ingestion/sources/registry.py` | 2008 | — (metadata-only) |
 | `cli/source_wizard.py` | 2288 | — |
-| `visualization/metabase.py` | 1152 | — |
+| `visualization/metabase.py` | 1151 | — |
 | `cli/init.py` | 1324 | — |
 | `visualization/dashboard_manager.py` | 1113 | — |
 | `cli/commands/platform.py` | 970 | — (extracted from main.py by TASK-005) |
--- a/dango/platform/common/startup.py
+++ b/dango/platform/common/startup.py
@@ -358,6 +358,9 @@ def _link_metabase_admin(project_root: Path, admin_email: str) -> None:
             dango_user.id,
             UserUpdate(metabase_user_id=mb_user["id"], metabase_password_enc=encrypted_pw),
         )
+        admin_creds["password"] = password
+        with open(mb_yml_path, "w") as f:
+            yaml.safe_dump(mb_creds, f, default_flow_style=False)
         _logger.info("metabase_admin_linked", email=admin_email, metabase_user_id=mb_user["id"])
     except Exception:
         # Non-critical — SSO bridge will lazy-sync on next login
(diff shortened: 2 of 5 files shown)


aaronteoh:
## Code Review — Approve ✓

### BUG-115: `startup.py` — Write new password back to `metabase.yml`

**Correctness: Good.** `admin_creds` is assigned as `mb_creds.get("admin", {})` at line 308. Since `dict.get()` returns the actual dict object (not a copy), mutating `admin_creds["password"]` updates `mb_creds["admin"]["password"]` in place. The `yaml.safe_dump(mb_creds, ...)` then writes the complete updated structure back.

**Ordering: Good.** The write happens after both `update_metabase_user_password()` (Metabase side) and `update_user()` (auth.db side) succeed. If either failed, the function would have already returned. The file write is the last step before the log line — correct position.

**Atomicity:** Non-atomic write is acceptable here — the entire function is wrapped in `except Exception: pass` (non-critical path), metabase.yml is ~5 lines, and worst case the SSO bridge lazy-syncs on next login. Atomic write-to-temp-then-rename would be overengineering.

**`default_flow_style=False`:** Matches how `setup_metabase` writes the file originally. Consistent.

### BUG-118: `metabase.py` — Hash-based container name

**Correctness: Good.** `DockerManager(project_root).compose_project_name` uses md5-hash-based naming (`dango-{hash}`), matching how `start_docker_services()` starts containers. The old `project_root.name` pattern would only work if the directory name happened to match Docker Compose's default, which it won't since `DockerManager._compose_env()` sets `COMPOSE_PROJECT_NAME` explicitly.

**Import placement: Good.** Lazy import inside function body — consistent with the existing `import subprocess` two lines above. No circular dependency risk (Level 2 → Level 2 same-level import).

**`DockerManager.__init__` side effects: None.** Constructor only sets `self.project_root` and `self.compose_file` (Path join). `compose_project_name` is a pure property (md5 hash). No I/O, no Docker daemon calls.

### Tests

- `test_happy_path_updates_metabase_yml` — verifies new password written to file AND matches what was sent to Metabase API. Good bidirectional check.
- `test_metabase_yml_unchanged_when_password_update_fails` — verifies negative path (400 response → file untouched).
- `test_uses_hash_based_container_name` — patches DockerManager at class level, verifies both `docker ps` filter and `docker restart` use hash-based name.
- `test_returns_false_when_container_not_running` — simple negative path.

### File exemptions

- `metabase.py`: 1152 → 1151 verified by `wc -l`. Correct.
- `test_platform_startup.py`: 601 lines, newly exempted. Correct.

### Verdict

Both fixes are minimal, correctly targeted, and well-tested. No security concerns, no architectural issues, no missing edge cases. All 3075 unit tests pass.

🤖 Review by Claude Code

PR #215, opened 2026-05-05
R10-B: CSV schema evolution NULL padding + zero-load error

aaronteoh:
## Summary

- **BUG-168 (Critical):** Replace `INSERT INTO target SELECT * FROM temp` with `_build_insert_select()` that explicitly lists columns with NULL padding for missing ones. Fixes Binder Error when CSV has fewer columns than target table.
- **BUG-169 (Medium):** Return `status: "error"` when all attempted file loads produce 0 rows (previously returned misleading "success"). Only fires when no deletions occurred.
- **BUG-170 (Medium):** Integration test proving `detect_table_drift()` correctly finds `column_added` after CSV schema evolution.

## Test plan

- [x] `pytest tests/unit/test_csv_schema_evolution.py -x` — 6 tests pass
- [x] `pytest tests/unit/test_csv_loader.py -x` — 22 tests pass
- [x] `pytest tests/unit/test_drift_detection.py -x` — 32 tests pass
- [x] `mypy dango/ingestion/csv_loader.py` — no issues
- [x] `pre-commit run --all-files` — all hooks pass
- [x] `wc -l dango/ingestion/csv_loader.py` — 922 lines (exemption updated)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -250,7 +250,7 @@ dango/                          # Python package source
 │
 ├── ingestion/                  # Level 1 — Data loading
 │   ├── dlt_runner.py           # ⚠ 2415 lines — orchestrates full sync pipeline
-│   ├── csv_loader.py           # Multi-format file loading with dedup (867 lines)
+│   ├── csv_loader.py           # Multi-format file loading with dedup (922 lines)
 │   ├── sources/
 │   │   └── registry.py         # Source metadata (33 source types)
 │   └── dlt_sources/            # ⚠ DO NOT MODIFY — vendored connectors (127 files)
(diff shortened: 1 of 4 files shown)


aaronteoh:
## Code Review: Approve ✓

### Correctness

**BUG-168 fix (`_build_insert_select`):** Sound approach. Iterates target columns in order, emits `"col"` for present columns and `NULL AS "col"` for absent ones — guarantees column-count and column-order alignment. The `_dango_*` metadata columns are present in both temp and target tables, so they match normally and never get NULL-padded. The `_dango_` prefix guard only suppresses metadata columns from the user-facing log.

**BUG-169 fix (zero-load detection):** Logic is sound. `stats["deleted"] == 0` guard avoids false-flagging deletion-only syncs. Partial success (some files load, some don't) correctly remains "success".

**BUG-170 test:** End-to-end integration with real DuckDB + SQLite (no mocks). Correctly demonstrates full lifecycle: load → baseline → evolve → drift detected.

### Observations (non-blocking, no changes needed)

1. Two extra DESCRIBE queries per file load — negligible for the intended use case (small number of files per sync).
2. `_build_insert_select` always called even when columns match — by design, degenerates to explicit `SELECT "col1", "col2", ...` which is functionally equivalent to `SELECT *`. More robust.
3. No SQL injection risk — column names come from DuckDB's own DESCRIBE output, quoted with double quotes.

### Test Coverage

- BUG-168: Asserts `total_rows == 3` and queries the actual NULL value
- BUG-169: Two tests cover both branches (empty-file error + deletion-only success)
- BUG-170: Real integration test with actual drift detection
- All existing tests (22 csv_loader + 32 drift_detection) remain green

LGTM — ship it.

PR #216, opened 2026-05-05
R10-D1: Fix notebook DuckDB connection + add guidance cells

aaronteoh:
## Summary
- **BUG-178 (Critical):** Replace `read_only=True` with `config={"access_mode": "read_only"}` in all three notebook templates (`explore.py`, `quality.py`, `blank.py`) to prevent WAL write lock that blocks `dango sync`
- **BUG-184 (Low):** Add `mo.md()` guidance cells to `explore.py` and `quality.py` explaining what the notebook does and how to interact with it

## Test plan
- [x] `grep -n "read_only=True"` returns no matches in templates
- [x] `grep -n "access_mode"` confirms all 3 files use config dict
- [x] `ruff check` + `ruff format --check` pass
- [x] `mypy` passes (no issues in 4 source files)
- [x] `pytest tests/unit/test_notebook_cli.py tests/unit/test_web_notebooks.py` — 42 tests pass
- [x] All files under 500 lines

### DIFF
--- a/dango/notebooks/CLAUDE.md
+++ b/dango/notebooks/CLAUDE.md
@@ -14,9 +14,9 @@ Marimo notebook server lifecycle, DuckDB read-only snapshots, file-level locking
 | `locking.py` | ~249 | File-level notebook locking via `notebook_locks` table | `acquire_lock()`, `release_lock()`, `refresh_lock()`, `force_release_lock()`, `is_locked()`, `get_lock_info()`, `copy_locked_notebook()` |
 | `proxy.py` | ~186 | HTTP + WebSocket reverse proxy to Marimo | `proxy_to_marimo()`, `proxy_websocket_to_marimo()` |
 | `templates/__init__.py` | ~5 | Package marker | — |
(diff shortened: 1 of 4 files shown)


aaronteoh:
## Code Review: LGTM ✓

**BUG-178 fix is correct.** `config={"access_mode": "read_only"}` sets the mode before database initialization, preventing WAL file creation entirely — unlike `read_only=True` which still initializes WAL. This matters for notebooks where connections persist for the entire Marimo session.

**BUG-184 guidance cells look good.** Concise, actionable markdown. Cell-local `import marimo as mo` follows Marimo conventions. Return tuple format is correct. Blank template correctly omitted (intentionally minimal).

**Observation (non-blocking, follow-up):** ~35 other call sites in the codebase still use `read_only=True`. These are short-lived server-side connections (millisecond WAL lock window) with no reported contention. Recommend a low-priority codebase-wide cleanup as a separate task.

No issues with cell ordering, dependency graph, style, or typing.

PR #220, opened 2026-05-05
R10-E3: Sources UX fixes (BUG-159, BUG-160)

aaronteoh:
## Summary
- **BUG-159:** "View in Catalog" link now uses `?source=` parameter instead of `?model=`. Catalog page handles the new param by filtering to tables belonging to that source, with a banner and "Show all" button to clear the filter.
- **BUG-160:** Status pill click now opens a compact sync history popup (last 5 syncs: relative time, status badge, duration, rows) instead of the full detail modal. "View full details" link bridges to the full modal.

## Test plan
- [ ] Sources page → click source name → full detail popup with config, history, stats
- [ ] Sources page → click status pill → compact popup with last 5 syncs only
- [ ] Compact popup → "View full details" → opens full detail popup
- [ ] Source detail popup → "View in Catalog" → catalog shows filtered list of that source's tables
- [ ] Catalog → `?source=chess` URL → shows only that source's tables, "Show all" clears filter
- [ ] No JS console errors on sources or catalog pages

### DIFF
--- a/dango/web/static/js/app.js
+++ b/dango/web/static/js/app.js
@@ -2655,6 +2655,7 @@ window.closeUploadModal = closeUploadModal;
 window.openSourceDetail = openSourceDetail;
 window.closeSourceDetail = closeSourceDetail;
 window.openSyncHistory = openSyncHistory;
+window.closeSyncHistory = closeSyncHistory;
 window.runDbtModel = runDbtModel;
 window.refreshDbtModels = refreshDbtModels;
 window.switchTab = switchTab;
--- a/dango/web/static/js/app.js
+++ b/dango/web/static/js/app.js
@@ -2024,7 +2024,7 @@ ${details.config.description ? `<div><span class="font-semibold text-gray-700">N
         // Set catalog link
         const catalogLinkEl = document.getElementById('detail-catalog-link');
         if (catalogLinkEl) {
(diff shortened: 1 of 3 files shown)


aaronteoh:
## Code Review

### Issue Found & Fixed

**`closeSyncHistory` not exported to `window` scope** — The inline `onclick="closeSyncHistory()"` in `sources.html` would throw `ReferenceError` at runtime because the function was defined in module scope but not added to the `window` exports block. Fixed in a6cc271.

### Verified Correct

- **XSS safety:** `escapeHtml()` used on all user-facing values (`time`, `duration`, `rows`). `statusClass`/`statusLabel` derived from internal logic only.
- **Graceful degradation:** `openSyncHistory()` falls back to full detail modal if compact modal element is missing.
- **Catalog `sourceFilter` integration:** Clean state machine integration. `_skipPush` guard in "Show all" button prevents double pushState. `sourceFilter` cleared in `goToList()`. URL serialization/restoration handles `?source=` param correctly.
- **Event listener pattern:** `.onclick` assignment replaces (doesn't accumulate), so no leak concern.
- **Edge case (non-existent source filter):** Shows empty list with "No items match" + "Show all" recovery button. Acceptable UX.

LGTM after the export fix.

PR #230, opened 2026-05-06
R10-H: PII override migration to YAML, remove web mutation, structured data heuristic

aaronteoh:
## Summary

- **BUG-162:** Migrate PII overrides from SQLite to `.dango/pii-overrides.yml` (version-controllable). Automatic one-time migration from SQLite on first access. Atomic writes via tmp+rename.
- **BUG-161:** Remove PUT/DELETE `/api/governance/pii/override` endpoints — PII management is now CLI-only (`dango governance pii-set`/`pii-list`). Remove dismissPii() click handler from catalog badge; tooltip now shows CLI instruction.
- **BUG-185:** Add structured data heuristic to suppress false PERSON detections on columns with long structured values (avg length >100 chars + delimiters like `[`, `]`, `{`, `\n`) — fixes chess PGN false positives.
- Remove dead `GOVERNANCE_PII_OVERRIDE_SET`/`DELETED` audit events and `PiiOverrideRequest` model.
- Update file-exemptions: `pii_detector.py` 602→614 lines.

## Test plan

- [x] `pytest tests/unit/test_pii_overrides.py` — full rewrite with YAML + migration tests (13 tests)
- [x] `pytest tests/unit/test_pii_detection.py` — 4 new structured data heuristic tests
- [x] `pytest tests/unit/test_governance_api.py` — 2 new endpoint-removed tests
- [x] `pytest tests/unit/test_web_governance_overrides.py` — rewritten for removed endpoints
- [x] Full unit suite: 3267 passed, 0 failed
- [x] ruff check + ruff format + mypy clean
- [x] Pre-commit hooks pass
- [x] Rebased onto v1


aaronteoh:
## Independent Code Review

### Correctness
**LGTM.** All three bugs addressed cleanly and changes are internally consistent.

### Security
- YAML loaded with `safe_load` (no arbitrary code execution)
- Removed PUT/DELETE endpoints had `governance.manage` permission — no regression since they're fully removed
- GET endpoint correctly still requires `governance.view`

### Edge Cases Verified
- `_yaml_to_dict` defaults empty string for missing keys — sort degrades gracefully
- SQLite `NULL` reason flows through correctly as `None` (`reason: str | None`)
- `_maybe_migrate_from_sqlite` catches all exceptions internally — never crashes callers
- `get_overrides_for_table` never-fail pattern preserved (outer try/except)

### Structured Data Heuristic (BUG-185)
- Placement is correct: runs after detection aggregation but before match-ratio filter
- avg_len > 100 guard prevents false suppression on short name-like columns
- Only requires one value with one delimiter — intentionally permissive for the chess PGN use case

### Acceptable Tradeoffs (not bugs)
- Double write on first-use-set (migration creates YAML, then set saves again) — natural flow, rare one-time event
- No file locking on YAML read-modify-write — PII overrides are a rare manual CLI operation; old SQLite version had the same class of issue

### Tests
- 75 targeted tests pass (overrides, detection, governance API, web overrides)
- Full unit suite: 3267 passed, 0 failed
- Migration tests cover: SQLite→YAML, YAML-exists skip, empty SQLite, migration failure

**Approved.** Clean implementation, correct edge case handling, thorough test coverage.

PR #246, opened 2026-05-12
R12-J: Fix username enumeration via lockout (BUG-235)

aaronteoh:
## Summary

- Track failed login attempts by IP+email in a new `login_attempts` SQLite table (migration 005), so both existing and non-existent emails get identical lockout behaviour (HTTP 423 after 5 attempts from the same IP)
- All three login failure paths (unknown email, inactive user, wrong password) now record attempts and can trigger lockout — previously only the wrong-password path did
- Admin unlock (`dango auth unlock` + web UI) clears both user-table and IP-based lockouts

## Test plan

- [x] 11 new unit tests in `TestIPBasedLockout` (unknown/known email lockout, IP independence, reset/unlock/cleanup, backwards compat)
- [x] 3 new web auth tests in `TestLoginLockoutEnumeration` (unknown email → 423, inactive user → 423, identical response body)
- [x] All 3409 unit tests pass
- [x] Integration test (5/5 stages) passes
- [x] ruff + mypy clean on all changed files

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -341,7 +341,7 @@ Full exemption registry: [`docs/file-exemptions.yml`](docs/file-exemptions.yml)
 | `cli/init.py` | 1334 | — |
 | `visualization/dashboard_manager.py` | 1112 | — |
 | `cli/commands/platform.py` | 1038 | — (extracted from main.py by TASK-005) |
-| `web/routes/auth.py` | 852 | — (split evaluated in DOC-025: exempt, security-critical) |
+| `web/routes/auth.py` | 886 | — (split evaluated in DOC-025: exempt, security-critical) |
 | `cli/commands/oauth.py` | 813 | — (renamed from auth.py by TASK-093) |
 | `web/helpers.py` | 851 | — (extracted from app.py by TASK-085) |
 | `ingestion/csv_loader.py` | 922 | — |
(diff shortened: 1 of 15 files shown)


aaronteoh:
## Code Review — Approve

Reviewed the full diff (15 files, +749/-36) as an independent expert reviewer. Two review passes — first identified 3 issues, second verified all fixes.

### Security

- **Fix is correct:** `login_attempts` table keyed by SHA-256 of `ip:email` gives identical lockout behaviour for existing and non-existent emails. Prevents username enumeration via lockout-timing side channel.
- **Timing oracle preserved:** `verify_password("dummy", DUMMY_HASH)` still called on unknown-email and inactive-user paths.
- **No new information leakage:** 423 response body identical across all failure paths (verified by `test_login_lockout_response_identical`).
- **Hash choice appropriate:** SHA-256 with `usedforsecurity=False` for lookup key, not security.

### Issues found and fixed

1. **`_update_user_table_best_effort` exception guard** — wrapped in try/except so a SQLite error doesn't propagate as 500. The `login_attempts` row is already committed before this runs.
2. **Redundant `update_user()` removed** — admin unlock (web + CLI) was calling `update_user()` then `unlock_account()` which does the same thing. Now just `unlock_account()`.
3. **`cleanup_expired_login_attempts()` wired into lifespan** — runs on server startup, purges rows >24h old. Prevents unbounded table growth.

### Test coverage

- 11 new unit tests (`TestIPBasedLockout`): unknown/known email lockout, IP independence, email independence, check/reset/unlock with IP, expired lockout, backwards compat, cleanup
- 3 new web integration tests (`TestLoginLockoutEnumeration`): unknown email → 423, inactive user → 423, identical response body
- All 3409 unit tests pass, integration test 5/5 stages pass
- Ruff + mypy clean

### Line counts verified

| File | `wc -l` | Docs |
|------|---------|------|
| `web/routes/auth.py` | 886 | 886 |
| `web/routes/users.py` | 531 | 531 |
| `cli/commands/auth.py` | 660 | 660 |
| `auth/lockout.py` | 448 | 448 |

### Minor note (not blocking)

- No test for `client_ip=None` in the web handler (when `request.client` is `None`). Falls back to legacy behaviour. Low risk — only happens in unusual proxy setups.

LGTM ✅

PR #288, opened 2026-05-27
fix: sanitize project directory name in install scripts

aaronteoh:
## Summary
User input with spaces/uppercase (e.g., "My Analytics") is sanitized to lowercase-dashes (e.g., "my-analytics") before directory creation.

Previously the raw input was used for mkdir but the `cd` instruction shown at the end would fail because the shell splits on spaces.

### Files changed
- `install.sh` — sanitize with tr
- `install.ps1` — sanitize with PowerShell string ops

No PyPI version bump needed — install scripts are served from GitHub raw content.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -703,6 +703,9 @@ function Main {
                 break
             } while ($true)
 
+            # Sanitize: lowercase, spaces → dashes, remove special chars
+            $projectDir = ($projectDir.ToLower() -replace '\s+', '-' -replace '[^a-z0-9-]', '')
+
             Write-Host ""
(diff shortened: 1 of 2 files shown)


aaronteoh:
## Code Review

- [x] Sanitization logic correct: lowercase, spaces → dashes, strips special chars
- [x] Both install.sh and install.ps1 updated
- [x] Placed after input validation, before mkdir
- [x] Tested: "My Analytics" → "my-analytics", normal inputs pass through unchanged
- [x] All CI checks pass
- Minor: leading/trailing spaces produce leading/trailing dashes — acceptable edge case

LGTM.

PR #304, opened 2026-06-04
Fix R2-D: Cloud detection, upgrade handling, backup restore, schedule status

aaronteoh:
## Summary
- **D1-D2:** Remove misleading cloud info notes from `dango start` and `dango sync` (triggered by local cloud.yml, not meaningful)
- **D3-D6:** Gate cloud warnings (`dual_scheduler`, `cloud_security`, `cloud_schedule_conflict`, systemd check) on `is_running_on_cloud()` (env var) instead of `has_cloud_deployment()`/`is_cloud_mode()` (cloud.yml exists)
- **D7:** Accept PEP 440 pre-release versions in `validate_version_string()` using `packaging.version.Version`
- **D8:** Fix cloud upgrade migration command — `python -c` with direct import instead of nonexistent `python -m dango.migrations run`
- **D9:** Add `--pre` flag to pip install when current/target version is pre-release; `include_pre=True` for PyPI auto-detect. Uses `Version.is_prerelease` for robust detection.
- **D10:** Add `--from-local` flag to `dango remote backup restore` (upload local file via SFTP, restore on server)
- **D11:** Add scheduler loaded state to `dango schedule status` — queries web API for job count, shows mismatch warnings
- **D12:** Stale PID detection verified — existing code is correct, no changes needed
- **D13:** Fix scheduler reload destroying missed job state:
  - (a) Move `_log_missed_recovery()` after `start()` so persisted jobs are loaded first
  - (b) Compare trigger (cron + timezone) before remove+re-add — unchanged triggers preserve `next_run_time` for missed job catch-up
  - (c) Fix `skipped_disabled` count — was counting updated jobs as disabled
- **D14:** Record `transform_error` in sync history when coalesced dbt fails (matches dlt_runner behavior). Changed log level from WARNING to ERROR.

## Verification
- `pytest -x`: 3695 pass, 0 fail, 31 skipped
- D1: Removed cloud info note from `dango start`
- D2: Removed cloud info note from `dango sync`
- D3-D5: Cloud warnings gated on `is_running_on_cloud()` (env var only)
- D6: `dango status` systemd check gated on `is_running_on_cloud()`
- D7: `validate_version_string("1.0.0b4")` accepts pre-release
- D8: Migration command uses `python -c` instead of `python -m`
- D9: pip install adds `--pre` when current version is pre-release
- D10: `--from-local` flag added to `dango remote backup restore`
- D11: Schedule status queries web API for APScheduler state
- D12: Stale PID detection verified — existing code is correct
- D13: Scheduler reload preserves unchanged triggers; missed recovery logged after start
- D14: Coalesced dbt failure recorded in sync history + log event

Note: Live verification (CLI commands, cloud upgrade) done by coordinating chat on beta-1.


aaronteoh:
## Code Review Summary (3 rounds)

**Verdict: Approved** — all genuine issues from three independent review rounds have been addressed.

### Fixes applied from review feedback
- **Pre-release detection:** Replaced fragile substring matching with `packaging.version.Version.is_prerelease`
- **PyPI fallback:** `check_latest_pypi_version(include_pre=True)` falls through to stable version when releases dict is empty
- **D11 job count:** Removed misleading comparison against `job_count` (which includes internal scheduler jobs) — now shows scheduler state with YAML-based enabled count
- **SSH cleanup guard:** `finally` block in `_backup_restore_from_local` wraps `exec_command` in try-except so dead SSH doesn't mask original error
- **Test coverage added:** 9 new tests covering `--from-local` restore (5 cases including failure), pre-release upgrade `--pre` flag (3 cases), and timezone-change trigger comparison (1 case)

### Findings confirmed as non-issues
- `check_latest_pypi_version` fallthrough: `return` exits the function; fallthrough only when releases is empty (intentional)
- Shell injection in migration command: `_PROJECT_DIR` is a compile-time constant, not user input
- String-based trigger comparison: APScheduler 3.x has no `__eq__` on CronTrigger; `str()` is the standard approach, confirmed working via tests
- Startup timing race: `_scheduler.start()` completes synchronously before `_log_missed_recovery()`

### Final state
- 6 commits, 3704 tests passing, 0 failures
- 22 files changed, +430/-130 lines

PR #316, opened 2026-06-07
Fix Windows compatibility bugs in local code paths (W2-W7)

aaronteoh:
## Summary

- **W2**: Guard `fcntl` import crash in `jobs.py` — conditional `fcntl`/`msvcrt` import + guarded `flock()` calls (matches `dbt_lock.py` pattern)
- **W3**: Guard `SIGTERM` handler in `watcher_runner.py` — `hasattr(signal, "SIGTERM")` on both instances (SIGINT still works on all platforms)
- **W4**: Replace `os.kill(pid, 15)` with cross-platform `kill_process()` in `network.py` (reuses existing `utils/process.py`)
- **W5**: Platform-aware `/etc/hosts` path in `network.py` — uses `C:\Windows\System32\drivers\etc\hosts` on Windows
- **W6**: Add Windows timezone fallback in `schedule.py` — tries `tzlocal` (optional dep), then `time.tzname`, before falling back to UTC
- **W7**: Generate Python-based pre-push hook on Windows in `init.py` (bash not available natively)

W1 (`dlt_runner.py` SIGALRM) was already fixed — confirmed and skipped.

## Verification
- `pytest -x`: 3725 passed, 31 skipped, 0 failures
- `grep -rn "^import fcntl" dango/`: only in `cloud/scheduled_backup.py` (Linux-only, out of scope)
- `grep -rn "os.kill.*\d+)" dango/`: remaining uses are all `os.kill(pid, 0)` (process-existence check, works cross-platform)
- All modified modules import cleanly

### DIFF
--- a/dango/cli/commands/schedule.py
+++ b/dango/cli/commands/schedule.py
@@ -157,7 +157,7 @@ def _get_local_timezone() -> str:
             tz_name = get_localzone_name()
             if tz_name:
                 return tz_name
-        except (ImportError, Exception):
+        except Exception:
             pass
         try:
             import time as _time
--- a/dango/cli/commands/schedule.py
+++ b/dango/cli/commands/schedule.py
@@ -137,6 +137,7 @@ def _get_local_timezone() -> str:
     3. Falls back to "UTC"
     """
     import os
+    import sys
 
     # Method 1: /etc/localtime symlink (macOS/Linux)
     try:
@@ -147,6 +148,25 @@ def _get_local_timezone() -> str:
                 return iana
     except (OSError, ValueError):
         pass
+
+    # Method 1.5: Windows — try tzlocal (IANA names), then time.tzname
+    if sys.platform == "win32":
+        try:
+            from tzlocal import get_localzone_name
+
+            tz_name = get_localzone_name()
+            if tz_name:
+                return tz_name
+        except (ImportError, Exception):
+            pass
+        try:
+            import time as _time
+
+            tz_name = _time.tzname[0]
+            if tz_name:
+                return tz_name
+        except Exception:
+            pass
 
     # Method 2: tzinfo.key (Python 3.12+)
     try:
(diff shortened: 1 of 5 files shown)


aaronteoh:
## Independent Code Review (3 rounds)

**Round 1** identified 6 issues — all fixed in subsequent commits:

| # | Issue | Severity | Resolution |
|---|-------|----------|------------|
| 1 | `msvcrt.LK_NBLCK` is non-blocking; `fcntl.LOCK_EX` is blocking | Bug | Changed to `LK_LOCK` (blocking, 10s retry) |
| 2 | `_add_pending_dbt_source` opens `"a+"` (position at EOF) — `msvcrt.locking()` would lock wrong byte range, breaking mutual exclusion with `_consume_pending_dbt_sources` (opens `"r+"`, position at 0) | Bug | Added `f.seek(0)` before `msvcrt.locking()` |
| 3 | `hasattr(signal, "SIGTERM")` guard is a no-op — `SIGTERM` is always defined on all Python platforms including Windows | No-op | Reverted to original code |
| 4 | `kill_process()` kills children + waits 10s, bypassing nginx's graceful worker shutdown; also doesn't raise on failure (silent success) | Behavior change | Replaced with `psutil.Process(pid).terminate()` — exact match to original `os.kill(pid, 15)` semantics |
| 5 | `except (ImportError, Exception)` — `ImportError` is a subclass of `Exception`, making the tuple redundant | Cleanup | Simplified to `except Exception` |
| 6 | Dead `import sys` in generated Python pre-push hook | Dead code | Removed |

**Round 2** confirmed all 6 fixes correct, flagged one remaining note (W4 nginx child behavior) — resolved in round 3 commit.

**Round 3** — clean approval, no remaining issues.

**Test results:** 3725 passed, 31 skipped, 0 failures (verified across multiple runs pre- and post-rebase).

✅ Approved

PR #330, opened 2026-06-13
fix: wizard UX bugs (P2-2, P2-4, P2-6, P2-7, P8-3)

aaronteoh:
## Summary

Fixes 5 OAuth wizard UX bugs that cause confusing behavior when OAuth fails or is skipped:

- **P2-2**: Only show OAuth success message when actual auth tokens (refresh_token/access_token) exist, not just client credentials
- **P2-4/P2-6**: Return `"skipped"` from `_handle_oauth_setup()` on "Skip for now", prompt user to confirm before continuing setup (exits cleanly via `click.Abort()` on decline)
- **P2-7**: GA4 `start_date` parameter now uses `type: "date"` with `default_days_ago: 90` (matching all other sources), instead of `type: "string"` with `default: "90daysAgo"`
- **P8-3**: "Credential setup required" block only shown when source has `secrets_toml_template` — CSV/local_files no longer trigger it

## Files changed

- `dango/cli/source_wizard.py` — P2-2, P2-4, P2-6, P8-3
- `dango/ingestion/sources/registry.py` — P2-7
- `tests/unit/test_source_wizard.py` — New: 9 tests covering all 5 bugs

## Verification
- `pytest -x`: 3790 pass, 0 fail, 31 skipped
- `ruff check dango/`: clean
- `ruff format --check dango/`: clean
- Live OAuth wizard testing deferred to coordinating chat (per verification workflow)

### DIFF
--- a/dango/cli/source_wizard.py
+++ b/dango/cli/source_wizard.py
@@ -179,6 +179,13 @@ class SourceWizard:
                         continue
                     elif oauth_result == "cancel":
                         return False
+                    elif oauth_result == "skipped":
+                        console.print(
+                            "\n[yellow]⚠  No access token obtained — syncs will fail "
+                            "until you complete authentication.[/yellow]"
+                        )
+                        if not click.confirm("Continue setup anyway?", default=False):
+                            raise click.Abort()
 
                     state = "params"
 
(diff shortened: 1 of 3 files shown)


aaronteoh:
## Code Review — Approve

Reviewed the full diff across 3 rounds. All 5 bug fixes are correct, well-scoped, and backward-compatible.

### Changes verified

| Bug | Fix | Status |
|-----|-----|--------|
| P2-2 | Token existence check (`refresh_token`/`access_token`) before success message | ✅ |
| P2-4/P2-6 | `"skipped"` return value + `click.Abort()` on decline | ✅ |
| P2-7 | GA4 `start_date`: `type: "date"` + `default_days_ago: 90` | ✅ |
| P8-3 | Gate credential block on `secrets_toml_template` | ✅ |
| (bonus) | Red error only on true OAuth failure, not partial success | ✅ |

### Key correctness checks

- **Backward compat (P2-7):** Existing `"90daysAgo"` configs still work — `dlt_runner.py:1373` skips GA4 for `daysAgo` → ISO conversion. GA4 Data API accepts both formats.
- **No silent fall-through (P2-4):** `elif` chain at lines 176-188 handles all four return values (`"back"`, `"cancel"`, `"skipped"`, `None`) without gaps.
- **Error messaging (P2-2):** Three branches cover has-tokens, no-tokens, and no-cred-at-all. Red error only shows on true failure (`run_oauth_for_source` returned `False`).
- **P8-3 condition:** `secrets_template` checked in outer `if`, so CSV/local_files (which have `setup_guide` but no `secrets_toml_template`) correctly skip the block.

### Tests

11 tests covering all 5 bugs. Sentinel `AssertionError` in skip-decline test prevents false passes. Full suite: 3790 pass, 0 fail.

### Issues found and resolved during review

1. Stale docstring for `_handle_oauth_setup` — added `"skipped"` return value
2. Weak skip-decline test — replaced `assert_not_called()` with sentinel error
3. P8-3 tests only checked registry data — added condition-logic behavioral tests
4. Contradictory error on partial OAuth success — moved red error into `else` branch
5. Inconsistent emoji (`⚠` vs `⚠️`) — normalized

PR #336, opened 2026-06-13
fix: geo seed auto-provisioning, GA4 date cast, staging schema cleanup

aaronteoh:
## Summary

- **P3-1**: Add `_provision_geo_targets()` to `DbtModelGenerator` so existing Google Ads projects get `geo_targets.csv` on every sync, not just during `dango source add`
- **P7-1**: Cast GA4 `date` columns from `TIMESTAMP WITH TIME ZONE` to `DATE` in staging models using DuckDB's `SELECT * REPLACE` syntax. Detection runs automatically for `google_analytics` sources
- **P7-3**: Drop empty `raw_*_staging` schemas from DuckDB after successful sync (Metabase already hides them; this removes them from DuckDB too)

## Files changed

- `dango/transformation/generator.py` — new `_provision_geo_targets()` method, `date_cast_columns` parameter in `generate_staging_model()`, detection logic in `generate_all_models()`
- `dango/templates/dbt/staging_model.sql.j2` — conditional `SELECT * REPLACE` for date cast
- `dango/ingestion/dlt_runner.py` — staging schema cleanup block in `run_sync()`
- `tests/unit/test_data_fixes.py` — 12 unit tests covering all three fixes

## Verification

- `ruff check`: all checks passed
- `ruff format --check`: all files formatted
- `pytest -x`: 3793 passed, 31 skipped, 0 failed

## Test plan

- [ ] Verify geo_targets.csv is provisioned on sync for existing Google Ads projects
- [ ] Verify GA4 staging models contain `CAST(date AS DATE)` after sync
- [ ] Verify `raw_*_staging` schemas are dropped from DuckDB after sync

### DIFF
--- a/dango/ingestion/dlt_runner.py
+++ b/dango/ingestion/dlt_runner.py
@@ -2389,6 +2389,31 @@ def run_sync(
 
     console.print(f"{'=' * 60}\n")
 
+    # Clean up empty dlt staging schemas after successful sync
+    if success_sources:
+        try:
+            import duckdb as _duckdb
+
+            _conn = _duckdb.connect(str(project_root / "data" / "warehouse.duckdb"))
+            try:
+                _schemas = _conn.execute(
+                    "SELECT schema_name FROM information_schema.schemata "
+                    "WHERE schema_name LIKE '%\\_staging' ESCAPE '\\'"
+                ).fetchall()
+                _dropped = []
+                for (schema_name_val,) in _schemas:
+                    base_name = schema_name_val.removesuffix("_staging")
+                    source_name = base_name.removeprefix("raw_")
+                    if source_name in success_sources or base_name in success_sources:
+                        _conn.execute(f'DROP SCHEMA IF EXISTS "{schema_name_val}" CASCADE')
+                        _dropped.append(schema_name_val)
+                if _dropped:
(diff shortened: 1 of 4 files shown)


aaronteoh:
**Code Review (independent pass)**

Reviewed all three fixes (P3-1, P7-1, P7-3). Confirming:

**Correctness:**
- P3-1: `_provision_geo_targets()` mirrors `source_wizard.py` logic with `if not exists` guards. Hooked into `generate_all_models()` for google_ads sources. Provisioning is logged.
- P7-1: `date_cast_columns` parameter is optional (backward-compatible). Detection correctly scopes to `google_analytics` sources with `TIMESTAMP` date columns. Template uses DuckDB `SELECT * REPLACE` — correct syntax.
- P7-3: Cleanup block in `run_sync()` is the common exit point for both `_run_dlt_source()` and `_run_dlt_native_source()` — no duplication needed. Only drops schemas for successful sources. Wrapped in try/except (non-fatal).

**Two-sync-path verification:** Confirmed `run_sync()` sits above both sync methods as the common orchestrator. The cleanup runs after the entire source loop, covering both paths. No mirroring needed.

**Tests:** 12 unit tests covering all paths including edge cases (already-exists, wrong source type, already-DATE, failure path, end-to-end).

**Full suite:** 3,793 passed, 31 skipped, 0 failed.

PR #339, opened 2026-06-15
fix: fd 0 clobbering breaks all scheduled syncs after terminal close

aaronteoh:
## Summary

When the dango server runs daemonized (launching terminal closed), fd 0 (stdin) is closed. `launch_sync_subprocess()` calls `open(log_path, "w")` which gets fd 0 as the lowest available fd. Then `subprocess.Popen(stdin=subprocess.DEVNULL)` does `dup2(devnull_fd, 0)` which clobbers the log file handle before it can be duped to stdout. The child Python process crashes with "Bad file descriptor" at `init_sys_streams`.

This caused 126 consecutive scheduled sync failures on beta-1 since Jun 14.

## Fix

Ensure fds 0-2 are occupied before opening the log file by filling any missing standard fds with `/dev/null`.

## Verification
- `ruff check dango/platform/sync_process.py`: passed
- `ruff format --check dango/platform/sync_process.py`: formatted
- `pytest tests/unit/test_sync_process.py -x`: 38 passed
- Manual verification needed on beta-1: restart dango, wait for scheduled sync, confirm success

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/dango/platform/local/watcher_lifecycle.py
+++ b/dango/platform/local/watcher_lifecycle.py
@@ -91,6 +91,11 @@ def start_file_watcher(project_root: Path) -> int | None:
     log_file = project_root / ".dango" / "watcher.log"
     log_file.parent.mkdir(parents=True, exist_ok=True)
 
+    # Prevent fd 0 clobbering: see ensure_std_fds() docstring for details.
+    from dango.utils.process import ensure_std_fds
+
+    ensure_std_fds()
+
     try:
         # Open log file
         log_handle = open(log_file, "w")  # noqa: SIM115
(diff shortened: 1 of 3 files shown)


aaronteoh:
**Independent code review — approved.**

- `ensure_std_fds()` is the standard POSIX daemon fd-guard pattern, correctly placed in `dango/utils/process.py`
- Both call sites (`sync_process.py`, `watcher_lifecycle.py`) call it immediately before `open(log_file)` — correct placement
- Idempotent, negligible overhead in normal case (3 `fstat()` calls), thread-safe
- `process_manager.py` intentionally excluded — runs from CLI with valid fds, not from daemonized server
- Lazy import pattern consistent with codebase conventions
- 62 related tests pass

PR #351, opened 2026-06-24
fix: logging noise + sync log filenames (FUP-5, FUP-7, FUP-8, FUP-10)

aaronteoh:
## Summary
- **FUP-5:** Add `log_level="INFO"` to test `configure_logging()` calls so they pass regardless of `DANGO_LOG_LEVEL` env var
- **FUP-7:** Add `console_level` parameter to `configure_logging()`, defaulting to `WARNING`, so terminal output is clean during syncs (file handler keeps the full resolved level)
- **FUP-8:** Suppress Presidio NLP entity warnings (CARDINAL, PRODUCT, MONEY) via `ner_model_configuration.labels_to_ignore`
- **FUP-10:** Include source name and timestamp in sync log filenames (`sync_<source>_<YYYYMMDD_HHMMSS>.log`) instead of unreadable hex sync_id
- Also increase Marimo cold start poll timeout from 10s to 30s (`range(10)` → `range(30)`)

## Verification
- `ruff format --check dango/ tests/`: 434 files already formatted
- `DANGO_LOG_LEVEL=ERROR pytest -x` (skipping pre-existing `test_notebook_manager.py` failures): **3872 passed, 31 skipped**

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/dango/governance/pii_detector.py
+++ b/dango/governance/pii_detector.py
@@ -162,6 +162,9 @@ def _get_analyzer() -> Any | None:
             nlp_configuration={
                 "nlp_engine_name": "spacy",
                 "models": [{"lang_code": "en", "model_name": _SPACY_MODEL}],
+                "ner_model_configuration": {
+                    "labels_to_ignore": ["CARDINAL", "PRODUCT", "MONEY"],
+                },
             }
         )
         _analyzer = AnalyzerEngine(nlp_engine=provider.create_engine())
--- a/dango/logging.py
+++ b/dango/logging.py
@@ -129,6 +129,7 @@ def configure_logging(
     log_level: str | None = None,
     log_dir: Path | None = None,
     json_console: bool = False,
+    console_level: str | None = None,
 ) -> None:
     """Configure structured logging for Dango.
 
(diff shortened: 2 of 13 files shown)


aaronteoh:
## Independent Code Review

All 6 fixes are correct and well-scoped. No issues found in final pass.

**FUP-5** — Tests are env-var independent. `test_env_var_override` correctly excluded from the blanket fix. `test_default_level_is_info` uses `monkeypatch.delenv` to truly test the default.

**FUP-7** — `console_level` cleanly separates console vs file levels. `_resolve_log_level(console_level or "WARNING")` correctly prevents `DANGO_LOG_LEVEL` from leaking into console output.

**FUP-8** — `ner_model_configuration.labels_to_ignore` is the correct Presidio API for suppressing NLP entity type warnings at the engine level.

**FUP-10** — Human-readable log filenames with source name + UTC timestamp. Redundant lazy imports cleaned up. Test assertion updated with regex.

**FUP-13** — `timeAgoIso()` properly deduplicated into `js/utils.js`. `base.html` uses `static_url()`. `formatRelativeTime()` keeps its tooltip wrapper; call sites unchanged.

**FUP-15** — `while True` is correct: can't write a PID file for an unready server. Exits only on process death or successful response.

**Verification:** ruff check/formatted, 3872 tests pass with `DANGO_LOG_LEVEL=ERROR`. Two pre-existing `test_notebook_manager.py` failures unrelated.

Ship it.

PR #360, opened 2026-07-05
fix: schedule execution history error display (field name + wrapping)

aaronteoh:
## Summary
- Fix schedule execution history: error messages never displayed due to field name mismatch (`error_message` → `error`)
- Use existing `<pre class="log-message">` pattern from sources.html for error display and proper word-wrapping

## Verification
- `ruff check dango/`: All checks passed, 216 files already formatted
- `ruff format --check dango/`: All checks passed
- `pytest -x`: 3910 pass, 31 skipped, 0 fail
- Manual: expand a schedule with failed runs — errors now display and wrap correctly


aaronteoh:
## Code Review: APPROVE

### Claims verified (independent confirmation)

| Claim | Evidence |
|---|---|
| DB column is `error` | `migrations/scheduler/001_execution_history.py:29` — `error TEXT` |
| `_finish_record` uses `error` param | `history.py:160` — `error: str \| None = None` |
| `_row_to_dict` produces `error` key | `history.py:67` — `dict(row)` uses column names |
| `log-message` CSS exists | `main.css:182-192` — `pre-wrap` + `break-word`, monospace |
| `sources.html` uses same pattern | `sources.html:230` — `<pre class="log-message text-red-700">` |

### Correctness: PASS
- `run.error` matches the JSON response — column is `error`, `_row_to_dict` uses `dict(row)`.
- No remaining `error_message` references in `schedules.html` after the fix.

### Security: PASS
- `x-text` (not `x-html`) — rendered as `textContent`, no XSS vector.

### Design consistency: PASS
- Same `<pre class="log-message">` pattern as `sources.html:230`. No new CSS.

### Testing: PASS
- `ruff check` / `ruff format --check` — clean
- `pytest -x` — 3910 pass, 31 skipped, 0 fail

### Regression risk: NONE
- Template-only change, no backend logic touched.

### Nit
`max-w-xs` on the `<td>` is now less critical since `log-message` handles wrapping, but it's harmless — provides a useful column width constraint (~320px). Good as-is.

PR #362, opened 2026-07-05
fix: remove broken global sync indicator from nav bar

aaronteoh:
## Summary
- Removes the broken global sync indicator from the nav bar in `base.html`
- The indicator was WebSocket-only with no initial state API call or polling fallback, couldn't persist across navigation, and was nearly invisible (`bg-gray-700` on `bg-gray-800`)
- Per-page indicators (Schedules `runningJobs`, Sources sync status) already handle sync state correctly with polling on load
- Also removes the orphaned `sync_queued` event mapping — the indicator was its sole consumer

## Changes
- Removed `x-data="globalSyncIndicator()" x-init="init()"` attributes from nav div
- Removed the indicator HTML div (clock SVG, spinner SVG, status text)
- Removed the entire `globalSyncIndicator()` script block (66 lines)
- Removed `"lock_waiting": "sync_queued"` from `_phase_to_event()` in `sync_process.py` (zero consumers remain; `lock_waiting` phases now fall through to `sync_progress` default, which both `app.js` and `schedules.html` handle)
- Updated test: `lock_waiting` assertion moved from `test_existing_mappings_unchanged` to `test_unknown_phase_defaults_to_sync_progress`

## Verification
- Grep confirms no remaining `globalSyncIndicator` or `sync_queued` references in source
- `ruff check dango/ tests/` passes
- `ruff format --check dango/ tests/` passes
- `pytest -x`: 3910 passed, 31 skipped, 0 failures


aaronteoh:
## Review: Approve

Clean removal with no loose ends. Verified:

- **No remaining references** to `globalSyncIndicator`, `sync_queued`, `queuedCount`, `activeSyncCount`, or `_postSyncSources` anywhere in the repo
- **Per-page indicators unaffected** — `app.js` (Sources) and `schedules.html` (Schedules) continue to consume `sync_started`, `sync_progress`, `post_sync_started`, `post_sync_completed`, `sync_completed`, `sync_failed`
- **Orphaned `sync_queued` mapping cleaned up** — `lock_waiting` phases now fall through to `sync_progress` default, which both consumers handle. The mapping line in `_phase_to_event()` and the corresponding test assertion were both updated
- **No Alpine.js scope issues** — the removed `x-data` was on a child `<div>`, independent of the parent `<nav>` component
- **Redundant WebSocket connection removed** — the nav bar was opening its own WS connection separate from the page-level connections

3910 pass, 0 fail, ruff clean.

PR #365, opened 2026-07-06
fix: add per-table empty replace protection for multi-resource dlt sources

aaronteoh:
## Summary
- Adds `_get_source_table_rows()` method returning per-table row counts (`dict[str, int]`)
- Refactors `_get_source_total_rows()` to delegate to the new method (preserves `int | None` API)
- Adds post-sync per-table check in both `_run_dlt_source()` and `_run_dlt_native_source()`
- Detects when individual tables would be truncated even when source-level totals pass

## Context
dlt 1.28.0 changes `replace` from dataset-level to per-table truncation. The existing source-level 0-row check uses total row counts, so a multi-resource source that returns data for 5 tables but 0 for the 6th would pass the check, silently truncating the 6th table.

The per-table check is additive — the existing source-level check is preserved and runs first.

## Verification
- `ruff check dango/` — all checks passed
- `ruff format --check dango/` — 216 files already formatted
- `pytest tests/unit/test_empty_replace_protection.py -v` — 48 passed (6 new)
- `pytest -x` — 3916 passed, 31 skipped, 0 failures

### DIFF
--- a/dango/ingestion/dlt_runner.py
+++ b/dango/ingestion/dlt_runner.py
@@ -1365,6 +1365,7 @@ class DltPipelineRunner:
 
             # Per-table empty replace protection: detect if any individual table
             # would be truncated (some tables populated, others empty)
+            post_table_counts = None
             if (
                 pre_table_counts is not None
                 and (full_refresh or uses_replace_mode)
@@ -1400,7 +1401,10 @@ class DltPipelineRunner:
                         }
 
             # Check for row count anomalies
-            total_row_count = self._get_source_total_rows(source_name)
+            if post_table_counts is not None:
+                total_row_count = sum(post_table_counts.values()) if post_table_counts else 0
+            else:
+                total_row_count = self._get_source_total_rows(source_name)
             if total_row_count is not None:
                 stats["total_row_count"] = total_row_count
             anomaly = self._check_row_count_anomaly(source_name, total_rows=total_row_count)
--- a/dango/ingestion/dlt_runner.py
+++ b/dango/ingestion/dlt_runner.py
@@ -964,6 +964,7 @@ class DltPipelineRunner:
 
         # Always capture pre-refresh row count for empty replace protection
         pre_refresh_rows = self._get_source_total_rows(source_name)
+        pre_table_counts = self._get_source_table_rows(source_name)
 
         # Full refresh: drop pipeline state only (NOT schema)
         # dlt with write_disposition="replace" handles table replacement automatically.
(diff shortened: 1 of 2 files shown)


aaronteoh:
## Expert Code Review — Approved

### Summary

The change is sound and achieves the stated goal. The per-table check is correctly placed after the source-level guard, both parallel methods are mirrored, `None`-handling is proper, and tests cover both paths.

---

### Changes Reviewed (commits 4e06a94 + 00cca5c)

**`dango/ingestion/dlt_runner.py`:**
- New `_get_source_table_rows()` returns `dict[str, int] | None` with per-table row counts
- `_get_source_total_rows()` refactored to delegate (preserves `int | None` API)
- Per-table check in both `_run_dlt_source()` and `_run_dlt_native_source()` — identical logic, correctly mirrored
- `post_table_counts` hoisted and reused in anomaly check (avoids redundant query)

**`tests/unit/test_empty_replace_protection.py`:** 7 new tests (Tests 32-38) in `TestPerTableProtection`:
- Partial empty fails with table details
- All populated succeeds
- New-table-with-zero-rows succeeds (no false positive)
- None pre-counts succeeds (first sync, skip check)
- `--allow-empty-replace` bypasses
- Source-level check fires first when all tables empty
- Native source path covered

---

### What's Good

- **Correct ordering.** Source-level check (`rows_loaded == 0`) fires first and returns early — cannot collide with the per-table check.
- **`None` guards are thorough.** Both `pre_table_counts is not None` and `post_table_counts is not None` independently skip the check on error.
- **`post_table_counts.get(table_name, 0)` correctly handles table disappearance.** If a previously-tracked table vanishes entirely post-sync, it's caught as 0 rows.
- **New-table immunity.** `pre_count > 0` guard prevents false positives for tables not in `pre_table_counts`. Test 34 validates.
- **Both methods mirrored.** Identical per-table logic in `_run_dlt_source` and `_run_dlt_native_source`.
- **`_restore_dlt_state` called consistently.** Both failure paths restore state.
- **Error message is actionable.** Lists each truncated table with row counts and the `--allow-empty-replace` escape hatch.
- **Existing tests pass without modification.** The plan kept `_get_source_total_rows` as a separate call preserving all 42 existing mock points.

---

### Verification
- `ruff check dango/` — all checks passed
- `ruff format --check dango/` — 216 files already formatted
- `pytest tests/unit/test_empty_replace_protection.py -v` — 49 passed (7 new)
- `pytest -x` — 3917 passed, 31 skipped, 0 failures

**Verdict: Approve.**

PR #383, opened 2026-07-13
fix: narrow Metabase stop/start in CLI sync and initial deploy paths

aaronteoh:
## Summary
- Narrow Metabase stop/start in CLI sync, CLI dbt, and initial deploy paths
- All four callers now stop Metabase only during the write phase, not extract
- Dashboards stay available during all extract phases

## Verification
- pytest -x: 3977 pass, 0 fail, 31 skipped
- ruff format --check dango/: 220 files already formatted

### DIFF
--- a/dango/cli/commands/source.py
+++ b/dango/cli/commands/source.py
@@ -667,7 +667,6 @@ def sync(
     console.print("🍡 [bold]Syncing data...[/bold]")
     console.print()
 
-    _metabase_was_stopped = False
     try:
         # Get project context
         project_root = require_project_context(ctx)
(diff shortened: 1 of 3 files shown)


aaronteoh:
## Independent Code Review — Approve

All three call sites correctly narrowed. No blocking issues.

### `source.py`
Outer stop/start + finally restart removed. Delegates to `run_sync()` internals (`_load_with_lock()` + dbt block). Post-sync schema sync now runs on cloud too — correct since Metabase is restarted after the write phase.

### `transform.py`
Does not go through `run_sync()`, so a narrow stop/start wraps only `subprocess.run(cmd)`. `DbtLock` released immediately after the dbt subprocess, before `update_model_status()`/`update_model_schemas()` (neither touches DuckDB). Local `try/finally` guarantees Metabase restart on exception/keyboard interrupt. Outer `finally` is a safety net for the lock.

### `initial_sync.py`
Removed stop/start entirely. `run_sync()` handles lifecycle.

### Verification
- ruff check: pass
- ruff format --check: pass
- pytest -x: 3977 pass, 0 fail, 31 skipped

### Remaining call sites (`stop_metabase_for_writes`/`start_metabase_after_writes`)
Only in `dlt_runner.py` (`_load_with_lock` + dbt block) and `transform.py` (narrow subprocess wrap). All correct.

PR #385, opened 2026-07-13
fix: full refresh broken for merge-based dlt_native sources

aaronteoh:
## Summary
- Fix full refresh for merge-based dlt_native and dlt sources: drop destination schema before pipeline.drop()
- pipeline.drop() only clears state, not data — merge/append sources need explicit schema drop
- Replace-based sources unaffected (they overwrite tables on every run)
- Fix applied to both `_run_dlt_native_source()` and `_run_dlt_source()` (parallel methods)

## Verification
- pytest -x: 4008 pass, 0 fail
- ruff format --check dango/: 220 files already formatted

## Manual verification needed (coordinating chat)
- `dango sync <merge_source> --full-refresh` — row count increased, full historical reload
- Incremental sync after full refresh — correct, only new rows
- `dango sync <replace_source> --full-refresh` — unaffected
- `--allow-empty-replace` protection still works

### DIFF
--- a/dango/ingestion/dlt_runner.py
+++ b/dango/ingestion/dlt_runner.py
@@ -970,15 +970,28 @@ class DltPipelineRunner:
         pre_refresh_rows = self._get_source_total_rows(source_name)
         pre_table_counts = self._get_source_table_rows(source_name)
 
-        # Full refresh: drop pipeline state only (NOT schema)
-        # dlt with write_disposition="replace" handles table replacement automatically.
-        # Removing pre-load schema drop preserves data when 0 rows are returned.
+        # Full refresh: for merge/append sources, drop schema + state to force a
+        # true reload. Replace sources skip the schema drop since they overwrite
+        # tables on every run (preserving schema protects against 0-row results).
         if full_refresh:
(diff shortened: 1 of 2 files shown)


aaronteoh:
## Review: PR #385 — Fix full refresh for merge/append sources

**Verdict: Approve.** The fix is correct, well-scoped, and all review issues have been addressed.

### What changed
- `_run_dlt_native_source()` and `_run_dlt_source()`: before `pipeline.drop()`, drop the destination schema for non-replace sources so merge/append sources get a true full reload
- Updated two static assert tests and added two behavioral tests

### Verified
- `pipeline.drop()` only clears local state, not destination tables (confirmed in dlt source)
- `_detect_write_disposition()` correctly gates the schema drop to merge/append only
- Empty-replace protection intact: `pre_refresh_rows` captured before the drop
- Connection properly closed via `try/finally`
- Comment accurately describes both paths
- **4010 tests pass, 0 fail**

### Minor note (non-blocking)
Behavioral tests cover `_run_dlt_source` specifically; `TestDropSchemaPresent` parametrized tests verify both methods at the source-text level.

PR #388, opened 2026-07-29
fix: @dlt.source decorator crashes during custom source import inspection

aaronteoh:
## Summary
- Fix @dlt.source decorator crash during custom source import inspection
- Register module in sys.modules before exec so dlt's inspect.getmodule() finds it

## Root cause
_import_and_inspect() creates modules via importlib but never registers them
in sys.modules. When the @dlt.source decorator executes and calls
inspect.getmodule(f), it returns None because the module isn't in sys.modules.
dlt then crashes on None.__name__.

## Verification
- ruff format --check dango/: 221 files already formatted
- pytest -x: 4030 pass, 31 skip, 0 fail
- Manual: custom_sources/ @dlt.source imports without warning in dango logs

### DIFF
--- a/dango/ingestion/sources/custom_discovery.py
+++ b/dango/ingestion/sources/custom_discovery.py
@@ -187,7 +187,11 @@ def _import_and_inspect(
                 logger.warning("Could not create module spec for %s", module_name)
                 return result
             module = importlib.util.module_from_spec(spec)
-            spec.loader.exec_module(module)
+            sys.modules[module_name] = module
+            try:
+                spec.loader.exec_module(module)
+            finally:
+                sys.modules.pop(module_name, None)
         except Exception:
             logger.warning(
                 "Could not import %s for metadata extraction", module_name, exc_info=True


aaronteoh (commented):
## Code Review: PR #388

### Summary
One-line fix (plus cleanup) to `_import_and_inspect()` — registers the dynamically loaded module in `sys.modules` before `exec_module()` so dlt's `@dlt.source` decorator doesn't crash when it calls `inspect.getmodule(f)`.

### What I checked

**Correctness: ✅ Pass**
- The root cause is accurate: `importlib.util.module_from_spec()` creates a module object but does NOT register it in `sys.modules`. When `@dlt.source` executes during `exec_module()`, dlt calls `inspect.getmodule(f)` which searches `sys.modules` by identity — finds nothing — returns `None` — crashes on `None.__name__`.
- Adding `sys.modules[module_name] = module` before `exec_module()` fixes this. dlt's introspection now finds the module.

**Cleanup: ✅ Pass**
- `sys.modules.pop(module_name, None)` in `finally` ensures the temporary registration doesn't leak. `None` default avoids `KeyError` on double-cleanup.
- The local `module` variable still holds a reference after the pop, so lines 202-220 (`getattr`, `inspect.getdoc`, `inspect.signature`) work correctly — they don't need the `sys.modules` entry.

**Error handling: ✅ Pass**
- The existing broad `except Exception` (line 195) still catches exec failures. The `finally` block runs before the `except`, so cleanup happens in both success and failure paths.
- If `sys.modules[module_name] = module` itself somehow raises (extremely unlikely), the `except` on line 195 catches it. The `finally` won't have run, but since the assignment failed, there's nothing to clean up.

**Regression risk: ✅ Very low**
- Only affects the Phase 2 import path, which is already wrapped in `except Exception` with graceful fallback.
- Phase 1 (AST parsing) is completely unaffected — that remains the primary discovery mechanism.
- No other callers use this function.

### Observations (non-blocking)

1. **Test coverage:** No tests exist for `_import_and_inspect()`. The test file (`tests/unit/test_custom_discovery.py`) has 0 references to it. Adding a test would require mocking `importlib.util.spec_from_file_location` and `exec_module`, which is fragile and low-value given the `except Exception` safety net. Skipping tests is reasonable for this fix.

2. **Thread safety (theoretical):** If `_import_and_inspect` were ever called concurrently, the temporary `sys.modules` entry could be visible to other threads during `exec_module()`. Currently this function is called sequentially during discovery, so not a practical concern.

### Verdict
LGTM. The fix is minimal, correct, and well-contained. No changes requested.

PR #393, opened 2026-08-10
fix: S3 QG staleness detection + unknown sync toast

aaronteoh:
## Summary
- QG-005: Add schedule-aware staleness detection to source status (2x interval threshold)
- QG-007: Fix "Unknown sync completed" toast — add source key to schedule broadcast
- Stale only applies to scheduled sources; unscheduled and manual-only unaffected
- Complex/unresolvable crons fall back to 24h threshold

## Verification
- pytest -x: 4030 pass, 0 fail
- ruff check: All checks passed
- ruff format --check: 221 files already formatted

### DIFF
--- a/dango/config/schedules.py
+++ b/dango/config/schedules.py
@@ -297,7 +297,7 @@ def validate_schedules(
         for sched in schedules:
             if sched.type not in (ScheduleType.SYNC, ScheduleType.SYNC_ONLY):
                 continue
-            interval = _get_cron_interval_seconds(sched.cron)
+            interval = get_cron_interval_seconds(sched.cron)
             for src in sched.sources:
                 avg = average_durations.get(src)
                 if avg is not None and interval < avg * 0.8:
@@ -312,7 +312,7 @@ def validate_schedules(
     return errors, warnings
 
 
-def _get_cron_interval_seconds(cron_expr: str, sample_count: int = 5) -> float:
+def get_cron_interval_seconds(cron_expr: str, sample_count: int = 5) -> float:
     """Return the minimum interval between consecutive cron runs in seconds.
 
     Samples ``sample_count`` consecutive intervals and returns the smallest one.
--- a/dango/platform/scheduling/jobs.py
+++ b/dango/platform/scheduling/jobs.py
@@ -713,6 +713,7 @@ def _run_scheduled_sync_impl(schedule_name: str, sources: list[str], **kwargs: A
                     "event": "sync_completed",
                     "schedule": schedule_name,
                     "sources": source_names,
+                    "source": source_names[0] if len(source_names) == 1 else schedule_name,
                     "duration_seconds": round(elapsed, 2),
                     "timestamp": _ts(),
                 }
--- a/dango/web/helpers.py
+++ b/dango/web/helpers.py
@@ -905,6 +905,45 @@ async def get_platform_health_data():
     }
 
 
+def build_staleness_thresholds() -> dict[str, float | None]:
+    """Pre-compute staleness thresholds for all scheduled sources.
+
+    Loads schedules config once and returns a ``{source_name: threshold_hours}``
+    dict. Sources not in any enabled schedule with a cron are excluded from the
+    dict (they are never stale). Callers look up a source and treat a missing
+    key as "not scheduled, never stale."
+
+    Returns:
+        Dict mapping source names to their staleness threshold in hours
+        (2x the minimum cron interval). Sources not in any schedule are omitted.
+    """
+    from dango.config.schedules import get_cron_interval_seconds, load_schedules_config
+
+    try:
+        config = load_schedules_config(get_project_root())
+    except Exception:
+        return {}
+
+    # Collect all cron intervals per source across all enabled schedules
+    source_intervals: dict[str, list[float]] = {}
+    for schedule in config.schedules:
+        if not schedule.enabled or not schedule.cron:
+            continue
+        try:
+            interval = get_cron_interval_seconds(schedule.cron)
+        except Exception:
+            interval = None
+        if interval is None:
+            interval = 24 * 3600  # complex/unresolvable cron fallback
+        for src in schedule.sources:
+            source_intervals.setdefault(src, []).append(interval)
+
+    # 2x the minimum interval per source
+    return {
+        src: (min(intervals) / 3600) * 2 for src, intervals in source_intervals.items() if intervals
+    }
+
+
 def _get_staleness_threshold_hours(source_name: str) -> float | None:
     """Return the staleness threshold in hours for a scheduled source.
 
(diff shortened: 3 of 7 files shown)


aaronteoh (commented):
## Independent Code Review — PR #393

### Overview
Three well-scoped changes fixing two S3 quality gate bugs. The implementation is correct and handles the edge cases described in the spec. A few items worth addressing before merge.

---

### ✅ What's Good

**Correctness.** The staleness logic is sound: only scheduled sources with `status == "synced"` can become stale, preserving the priority chain (failed > stale > partial > not_synced > empty > synced). Manual-only schedules and unscheduled sources are correctly excluded.

**Edge case coverage.** Multi-schedule sources use the minimum interval. Complex crons fall back to 24h. The `schedule.cron` guard handles None/empty. All match the task spec.

**Defensive coding.** The `try/except` around `load_schedules_config` prevents config errors from crashing the status endpoint. The `try/except` around `_get_cron_interval_seconds` handles malformed crons gracefully.

**Broadcast fix is minimal.** The one-line addition in `jobs.py` is exactly what's needed — no scope creep.

---

### ⚠️ Concerns

**1. Repeated config file I/O (performance)**

`_get_staleness_threshold_hours` loads and parses `schedules.yml` from disk on every call. Since `get_source_status_data` runs per-source via `asyncio.gather`, a project with 10 sources will parse the same YAML file 10 times in rapid succession. For a small number of sources this is negligible, but it's unnecessary work.

*Suggestion:* Pass the pre-loaded schedules config into the function, or load it once in the route handler (similar to how `_build_schedule_map` works in `routes/sources.py`). Alternatively, accept it as-is and note it as a low-priority optimization — this isn't a correctness issue.

**2. No test coverage for the new logic**

`_get_staleness_threshold_hours` has 6+ distinct branches (unscheduled, manual-only, normal cron, complex cron fallback, multi-schedule minimum, config load failure) but none are tested. The `get_source_status_data` staleness path is also untested. While `pytest -x` passes, it's only verifying that nothing is broken — not that the new code is correct.

*Suggestion:* At minimum, add unit tests for `_get_staleness_threshold_hours` covering:
- Returns `None` for unscheduled source
- Returns `None` when all schedules are manual-only
- Returns `2 * interval_hours` for a simple cron (e.g., daily = 48h)
- Returns `24 * 2 = 48h` when cron is unresolvable
- Returns shortest interval when source is in multiple schedules
- Returns `None` when `load_schedules_config` raises

**3. Private import across modules**

`_get_cron_interval_seconds` is a private (`_`-prefixed) function in `dango.config.schedules`. Importing it from `dango.web.helpers` couples the web layer to an internal implementation detail of the config layer. If the config module renames or changes this function's signature, the web layer breaks silently.

*Suggestion:* Either (a) make `_get_cron_interval_seconds` public (drop the `_` prefix) since it's now used by two modules, or (b) add a public wrapper in `config/schedules.py` that exposes the interval calculation. The function is already well-defined and stable — making it public is the pragmatic choice.

**4. `sourceStatusOrder` behavioral note (not a bug)**

The updated function checks `order[src.status]` before `order[src.freshness?.status]`. For existing statuses this produces identical results (e.g., `src.status = "failed"` → 50, same as `order["failed"]`). The only new behavior is `src.status = "stale"` → 15, which is the intended case. This is correct but worth noting that it subtly broadens the function's input surface — if `src.status` ever contains a value not in the order map, it falls through to the freshness check (the old behavior). No issue in practice.

---

### 📋 Summary

| Category | Assessment |
|----------|-----------|
| Correctness | ✅ Correct |
| Edge cases | ✅ Covered |
| Performance | ⚠️ Repeated YAML parsing (minor) |
| Test coverage | ⚠️ No tests for new logic |
| Module boundaries | ⚠️ Private import across layers |
| Regression risk | ✅ Low — status field only added, never removed |

**Verdict: Approve with suggestions.** All concerns are non-blocking. The private import and missing tests are worth addressing but don't need to hold up merge for an S3 quality gate fix.

aaronteoh (commented):
## Follow-up Review — PR #393 (revised)

Re-reviewing after the three fixes from the initial review.

---

### ✅ Previously Identified Issues — All Resolved

**1. Private import** (fixed)
`get_cron_interval_seconds` is now public. The rename is clean — definition, both callers in `helpers.py`, and the two existing tests in `test_schedules_config.py` all updated correctly.

**2. Repeated config file I/O** (fixed)
`build_staleness_thresholds()` loads schedules.yml once. The route handler in `sources.py` calls it via `asyncio.to_thread` before `asyncio.gather`, then passes the dict to each `get_source_status_data` call. The new `staleness_thresholds` kwarg is optional with a `None` default — backward compatible if `get_source_status_data` ever gets another caller.

**3. Test coverage** (fixed)
11 tests, all passing. Coverage includes the fallback path (`_get_staleness_threshold_hours`) and the optimized path (`build_staleness_thresholds`). Config load failures, unscheduled sources, disabled schedules, unresolvable crons, cron exceptions, and multi-schedule minimums all tested.

---

### 🔍 New Observations

**Minor: Logic duplication between `build_staleness_thresholds` and `_get_staleness_threshold_hours`**

Both functions contain essentially identical logic (iterate schedules, filter disabled/no-cron, call `get_cron_interval_seconds`, collect intervals, return 2x minimum). The only difference is one builds a dict for all sources while the other computes for one.

Not a blocker. The duplication is intentional — `_get_staleness_threshold_hours` exists as a standalone fallback when no pre-computed dict is available (e.g., a future caller that doesn't have the dict). It would be worse to have it call `build_staleness_thresholds` and look up one key, since that loads all schedules for a single source.

*Verdict: Acceptable duplication. Could add a `# Keep in sync with build_staleness_thresholds()` comment on both functions as a future-proofing hint.*

**Minor: `build_staleness_thresholds` return type is slightly misleading**

The return type `dict[str, float | None]` suggests values can be `None`, but the implementation never puts `None` in the dict — sources with no intervals are simply excluded. The type annotation is correct for `get_source_status_data`'s `.get(source_name)` pattern (missing key returns `None`), but the function itself always returns `dict[str, float]`.

*Verdict: Not a bug. The annotation matches how the dict is consumed. Could tighten to `dict[str, float]` for clarity.*

---

### 📋 Final Verdict

| Category | Before | After |
|----------|--------|-------|
| Correctness | ✅ | ✅ |
| Edge cases | ✅ | ✅ |
| Performance | ⚠️ Repeated YAML I/O | ✅ Single load |
| Test coverage | ⚠️ None | ✅ 11 tests |
| Module boundaries | ⚠️ Private import | ✅ Public function |
| Regression risk | ✅ Low | ✅ Low |

**Verdict: Approved.** All issues resolved. The two new observations are cosmetic and don't warrant changes.

aaronteoh:
Approved. All three review findings resolved: public function rename, single config load per request, 11 unit tests. 4041 pass, 0 fail.

PR #394, opened 2026-08-10
fix: S3 QG polish — column alignment + underscore-name empty-replace edge case

aaronteoh:
## Summary
- QG-003: Align sources table columns consistently, move Actions to leftmost
- QG-004: Fix empty-replace protection for underscore-prefixed source names

## Verification
- pytest -x: 4041 pass, 0 fail
- ruff check / format --check: all pass

### DIFF
--- a/dango/ingestion/dlt_runner.py
+++ b/dango/ingestion/dlt_runner.py
@@ -988,8 +988,12 @@ class DltPipelineRunner:
         state_backup = self._backup_dlt_state(pipeline_name)
 
         # Always capture pre-refresh row count for empty replace protection
-        pre_refresh_rows = self._get_source_total_rows(source_name)
-        pre_table_counts = self._get_source_table_rows(source_name)
+        pre_refresh_rows = self._get_source_total_rows(
+            source_name, schema_name=pipeline.dataset_name
+        )
+        pre_table_counts = self._get_source_table_rows(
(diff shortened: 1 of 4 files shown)


aaronteoh:
## Review: PR #394 — S3 QG Polish

### `dlt_runner.py` — QG-004

- `_get_source_table_rows` / `_get_source_total_rows`: optional `schema_name` param with `or` fallback is the right pattern — backward-compatible, handles both `None` and `""`.
- All 6 call sites across both `_run_dlt_source` and `_run_dlt_native_source` updated. Verified `pipeline` is in scope at every call site.
- Correctly left `_get_csv_table_rows` (no dlt normalization) and `_check_row_count_anomaly` (advisory, returns None on failure) unchanged.

### `sources.html` + `app.js` — QG-003

- `<th>` and `<td>` alignment classes now consistent: `text-left` on text columns, `text-right` on Rows.
- Column count preserved (7), `colspan="7"` in empty states still correct.
- Sort is key-based (`data-sort-key`), filter is text-based — column reorder doesn't affect either.

### `test_dlt_runner_pipeline_phases.py`

- One-line fix: `p.dataset_name = "raw_test_source"` on the spec'd mock. Other test files use plain `MagicMock()` so they auto-create the attribute.

### Verdict

Approve. 4041 tests pass, ruff clean.

PR #395, opened 2026-08-10
fix: replace nth-child with data-column attribute for status cell

aaronteoh:
## Summary
- Replace fragile `td:nth-child(2)` with `td[data-column="status"]` in `updateDbtModelStatus()`
- Add `data-column="status"` to status cell in `renderDbtModelsTable()`

## Verification
- `pytest -x`: 4041 pass, 0 fail
- Manual: Run button places running pill in correct column

### DIFF
--- a/dango/web/static/js/app.js
+++ b/dango/web/static/js/app.js
@@ -2703,7 +2703,7 @@ function renderDbtModelsTable() {
                 <td class="px-6 py-4 whitespace-nowrap">
                     <div class="text-sm font-medium text-gray-900">${escapeHtml(model.name)}</div>
                 </td>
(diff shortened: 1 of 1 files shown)


aaronteoh:
Approve. Clean two-line fix.

- `nth-child(2)` fully removed from app.js (zero remaining)
- `data-column="status"` is unique — no naming collisions in JS or templates
- Scoping is safe: `closest('tr')` from `dbt-btn-${modelName}` always returns a models table row
- `statusBadge` values are all `<span>` elements — can't interfere with the `td[...]` selector
- 4041 pass, 0 fail

No changes requested.

PR #400, opened 2026-08-16
chore: guard against stale compiled Tailwind CSS

aaronteoh:
## Summary
- `make css-build` now appends a trailing newline, so it stops tripping `end-of-file-fixer` on every rebuild.
- Add a CI job that rebuilds the CSS from templates/JS and fails on `git diff` drift.

## Why
The compiled `tailwind.min.css` was chronically stale: templates/JS referenced utilities (`rounded-t-lg`, `select-none`, `italic`, `text-blue-500`, etc.) that were missing from the compiled file, silently breaking styles. `web/CLAUDE.md` says "rebuild after template changes" but nothing enforced it. This guard stops that class of bug permanently.

## Verification
- `make css-build` in a clean worktree → `git diff --exit-code` returns 0 (rebuilt CSS matches committed byte-for-byte, trailing newline included).
- Negative test: appended a fake class, `git diff --exit-code` returned 1 (guard catches drift).
- Pre-commit hooks pass (end-of-file-fixer, check-yaml, etc.).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,20 @@ on:
     branches: [v1, main, 'v1.0.*']
 
 jobs:
+  css:
+    name: CSS build (tailwind)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+          cache: npm
+      - name: Build CSS
+        run: make css-build
+      - name: Check no drift
+        run: git diff --exit-code dango/web/static/css/tailwind.min.css
+
   lint:
     name: Lint (ruff)
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


aaronteoh:
## Code review — approved

Correct and minimal. Verified the two coupled concerns are handled in the right order:

1. **Trailing-newline fix** — `--minify` emits no newline (byte-verified: rebuilt 31928B ending in `}` vs committed 31929B ending in `\n`), so `end-of-file-fixer` was tripping on every rebuild. The `printf '\n'` resolves it, and the rebuild is now byte-identical to committed.
2. **Drift guard** — `make css-build` runs before `git diff --exit-code`, comparing fresh output against the committed artifact. Negative-tested: injected a fake class → exit 1 (caught).

Findings (none blocking):
- **Scope is correct** — only `tailwind.min.css` is compiled; `main.css`/`catalog.css` are handwritten and correctly excluded.
- **Node 20 vs 24** — Tailwind 3.4 output is Node-version-agnostic, so no false drift today. Optional future hedge: align `node-version` or add an `engines` field. Not required.
- **CI confirms** — the new `css` job passed (16s), so the `git diff --exit-code` semantics work in GitHub Actions as expected.

Merge-ready.

PR #424, opened 2026-08-21
1.0.7-13: Fix schedule reload ignoring source-list changes

aaronteoh:
## Summary

Fixed `reload_schedules()` to detect source-list and timeout modifications by adding kwargs diffing alongside trigger diffing. Previously, the function only compared cron triggers, causing source additions/removals and timeout changes to be silently ignored across restarts.

**Real incident on beta-1:** `negative_keywords` and `change_history` sources were added to `badang_labs_sem_daily_sync` in YAML without cron changes. The old pickled job definition stayed active, silently blocking the new sources for 6+ days.

Also fixes service account validation to return `valid=False` when `google-auth` is not installed, surfacing configuration errors in health checks.

### `dango/config/schedules.py`
- Added `_RELOAD_KWARGS_EXCLUDE = frozenset({"_timeout_minutes"})` constant (line 63)
- Replaced `if job_id in existing_ids:` block (596–630) with kwargs diffing logic:
  - Compute `kwargs_changed` by comparing filtered kwargs dicts (excluding `_timeout_minutes`)
  - Remove + re-add job when `trigger_changed or kwargs_changed`
  - Preserve `next_run_time` only when kwargs changed (trigger didn't)
  - Drop `next_run_time` when trigger also changed (old fire time invalid with new cron)

### `dango/oauth/service_account.py`
- Changed line 220: `valid=True` → `valid=False` when `google-auth` not installed
- Updated message to include helpful install command
- Ensures credentials are flagged as invalid when verification is impossible

### `tests/unit/test_schedules_loading.py`
- Updated `_make_scheduler()` to support 3-tuple job definitions with kwargs/next_run_time
- Updated `test_unchanged_trigger_preserves_job` to pass matching kwargs
- Added 3 new tests: `test_changed_sources_updates_job`, `test_timeout_change_only_stays_unchanged`, `test_kwargs_and_trigger_both_changed_updates_job`

## Verification

- `pytest tests/unit/test_schedules_loading.py::TestReloadSchedules -v`: All 11 tests pass ✓
- `ruff check dango/`: No issues ✓
- Code review: Zero findings ✓

## Design Notes

- **`_timeout_minutes` exclusion:** Old persisted jobs don't have this field. Including it would mark every legacy job as changed on first reload. Exclusion preserves first-reload compatibility.
- **`next_run_time` preservation:** Only when kwargs change but trigger stays the same, preventing catch-up scheduling from being destroyed.
- **Service account fix:** Returns `False` when google-auth missing (not transient). More useful than silent benefit-of-the-doubt for a missing library.

### DIFF
--- a/dango/config/schedules.py
+++ b/dango/config/schedules.py
@@ -60,6 +60,8 @@ _SLUG_RE = re.compile(r"^[a-z][a-z0-9_]*$")
 
 _VALID_NOTIFY_ON = frozenset({"failure", "success", "stale"})
 
+_RELOAD_KWARGS_EXCLUDE = frozenset({"_timeout_minutes"})
+
 
 def get_schedule_job_id(schedule_name: str) -> str:
     """Return the APScheduler job ID for a schedule name.
@@ -594,7 +596,7 @@ def reload_schedules(
             }
 
         if job_id in existing_ids:
-            # Check if trigger actually changed before removing.
+            # Check if trigger and kwargs actually changed before removing.
             # Removing and re-adding destroys APScheduler's stored next_run_time,
             # which prevents catch-up of missed jobs on restart.
             existing_job = existing_jobs[job_id]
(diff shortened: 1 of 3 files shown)


aaronteoh:
✅ Code review approved. All findings addressed:
- schedules.py kwargs diffing: correct
- service_account.py valid=False: correct  
- Tests: all 4 correct with next_run_time verification
- scheduler.py auto-reload: removed (out-of-scope)

Ready to merge.

PR #414, opened 2026-08-20
1.0.7-1: Model wizard table selection + CTE scaffold

aaronteoh:
## Summary

Adds interactive upstream table selection to `dango model add` command with automatic CTE scaffold generation:
- Intermediate models limited to staging layer; marts models see all layers
- Looping single-select prompt (never checkbox) for table selection
- Generated CTE scaffold with alias extraction from table names
- Fallback to placeholder SQL when no tables selected
- All CTEs reference upstream via `{{ ref() }}`

## Changes

**dango/cli/model_wizard.py** (507→547 lines)
- `_get_available_tables()` — discovers staging/intermediate/marts models
- `_ask_upstream_tables()` — looping single-select prompt
- Updated `_generate_sql_template()` to generate CTEs instead of 60-line reference guide
- Updated `_create_model_file()` to call table selection

**tests/unit/test_model_wizard.py** (new, 212 lines)
- 10 tests covering CTE scaffold generation, table discovery, layer filtering, alias derivation

**Documentation**
- Updated model_wizard.py line count (507→547) in 3 files:
  - docs/file-exemptions.yml
  - dango/cli/CLAUDE.md  
  - CLAUDE.md (workspace)

## Test Plan

- [x] pytest tests/unit/test_model_wizard.py — 10/10 pass
- [x] ruff check — all pass
- [x] ruff format — clean
- [x] Pre-commit hooks — all pass
- [ ] Manual test on beta-1 (coordinating chat)

## Verification

```bash
# Unit tests
pytest tests/unit/test_model_wizard.py -v
# 10 tests: CTE scaffold (2 with/without tables, 1 empty list), 
# table discovery (empty project, with models, all layers, layer filters),
# alias derivation (single and multiple tables)

# Code quality
ruff check dango/cli/model_wizard.py tests/unit/test_model_wizard.py
ruff format --check dango/cli/model_wizard.py tests/unit/test_model_wizard.py
```

Manual testing deferred to coordinating chat (beta-1 verification workflow).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Myfp1v1S8UKqXEnZ2iTuq4

PR #426, opened 2026-08-21
fix: model wizard duplicate CTE alias on double collision

aaronteoh:
## Summary
- Fix model wizard generating duplicate CTE names when full table name collides with an already-used alias (double collision case not handled by PR #414)
- Example: `["stg_stripe_customers", "customers"]` previously generated two CTEs both named `"customers"` → invalid SQL
- Fix: add numeric suffix (`customers_2`, `customers_3`, …) when fallback also collides

## Changes
- `dango/cli/model_wizard.py`: Added second collision check with numeric suffix generation (lines 484-487)
- `tests/unit/test_model_wizard.py`: Added `test_alias_double_collision_resolution` test case

## Verification
- `pytest -x`: 4199 pass, 30 skip
- `test_alias_double_collision_resolution`: new test covers the double collision case
- `test_alias_collision_resolution` (PR #414 case): passes unchanged ✓
- All other alias tests: pass unchanged ✓
- Full test suite: all green

## Regression check
- Single collision case (stg_stripe_customers, stg_hubspot_customers) — no numeric suffix needed, falls back to full table name as before
- No collisions case — new code path never triggers
- All existing tests pass unchanged

### DIFF
--- a/dango/cli/model_wizard.py
+++ b/dango/cli/model_wizard.py
@@ -480,6 +480,12 @@ class ModelWizard:
                 # Fall back to full table name if alias already used
                 if alias in aliases:
                     alias = table
+                # If fallback also collides, add numeric suffix
+                if alias in aliases:
+                    i = 2
+                    while f"{alias}_{i}" in aliases:
+                        i += 1
+                    alias = f"{alias}_{i}"
                 aliases.append(alias)
 
             # Generate CTE block with resolved aliases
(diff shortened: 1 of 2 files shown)


aaronteoh:
## ✅ Independent Code Review: APPROVED

I've completed a thorough independent code review covering all critical angles:

**Verification:**
- ✓ Logic verified: derived alias → fallback → numeric suffix (correct)
- ✓ Original PR #414 case: single collision still works unchanged
- ✓ Test coverage: new `test_alias_double_collision_resolution` reproduces exact bug scenario
- ✓ All 4199 tests pass, zero regressions
- ✓ Code follows repository patterns and conventions

**Implementation quality:**
- Root cause addressed properly (not a bandaid)
- Minimal scope (6 lines)
- O(1) complexity appropriate for non-hot-path wizard code
- Edge cases handled: no collisions, single collision, double collision all work

**Verdict:** Production-ready. No issues found.

PR #428, opened 2026-08-26
fix: scheduled script runs now appear in scripts UI history

aaronteoh:
## Summary
- Fix scheduled script runs not appearing in scripts UI history (last_run always null)
- Root cause: _write_script_logs() in jobs.py wrote to timestamped directories but not
  to the .jsonl file that the scripts API reads for last_run
- Fix: append to .jsonl at end of _write_script_logs(), matching path from scripts_helpers

## Follow-up fix #1 (after independent review)
An independent review of the first commit found the entry schema and log directory layout
didn't actually match what the UI needs:
- Missing `run_id`/`started_at` meant `last_run` stopped being null but rendered no usable
  timestamp (`—`), history rows sorted incorrectly, and the history row key was `undefined`.
- Log files were written to `scripts/{name}_{timestamp}/`, but the log viewer route reads
  `scripts/runs/{run_id}/` — so "View Log" would 404 for every scheduled run.

Second commit threads a `run_id` (uuid4) and `started_at` through `run_scheduled_script()` →
`_write_script_logs()`, writes stdout/stderr/meta.json under `scripts/runs/{run_id}/` to
match `_get_log_dir()`, and includes `run_id`/`started_at`/`error` in the jsonl entry to
match the manual-run schema in `scripts_api.py`.

## Follow-up fix #2 (after a second, scoped independent review)
A second review of the updated diff found two real issues in the new code:
- The failure-path `_write_script_logs()` call wrote a generic `"Exit code: N"` error to
  the jsonl entry, while activity log / notifications / execution history used a richer
  stderr-derived message computed a few lines later in the same function — so the Scripts
  UI showed a worse error than every other surface. Fixed by computing the message once
  and reusing it.
- Dropped an unnecessary local re-import/alias of `datetime`/`timezone` inside
  `_write_script_logs` (both already imported at module level; the `started_at: datetime`
  annotation never evaluates at runtime since this file uses
  `from __future__ import annotations`).

## Verification
- `pytest -x`: 4209 passed, 30 skipped, 0 failed
- `pytest tests/unit/test_script_history.py -x -v`: 5 passed (incl. a test asserting
  stdout/meta land under `_get_log_dir()/{run_id}`, matching the web log-viewer route)

## Regression check
- Both _write_script_logs call sites (timeout + success/failure) updated together
- jsonl append and log-dir write are inside existing try/except — failure is logged at
  DEBUG, never raised
- _load_history() reads jsonl in reverse order — appended entries are read correctly
- No other code references the old `scripts/{name}_{timestamp}/` log directory layout
  (grepped before changing it)

## Known issues NOT fixed here (flagged across two independent reviews, deliberately deferred)
- **Schema duplication across writers**: the jsonl entry schema is hand-duplicated between
  `jobs.py` and `scripts_api.py` with no shared type — the exact class of bug that caused
  the first-commit gap. Not fixed here since it'd require a shared module touching both
  write paths, beyond this bug's scope. Worth a follow-up if a third writer ever appears.
- **Unlocked concurrent jsonl writer**: both writers use unlocked `open(..., "a")`. This PR
  adds a second concurrent writer to a file that already had no locking for the manual-run
  path. Judged not a real risk (single short line, POSIX O_APPEND) and consistent with
  existing precedent.
- **Cancellation / pre-launch-exception paths still write no per-run history**: `except
  JobCancelledError`, `except JobTimeoutError` (outer resilience-layer timeout, distinct
  from the inner subprocess timeout this PR already handles), and `except Exception` in
  `run_scheduled_script` never called `_write_script_logs` even before this PR — confirmed
  via `git diff origin/v1.0.7...HEAD`, this PR did not touch those paths. A scheduled run
  that's cancelled via the UI, or that fails before the subprocess even launches, still
  leaves no history entry. This is a real, pre-existing gap, but closing it is new
  functionality (e.g. deciding what a "cancelled before it started" history entry looks
  like with no stdout/stderr/exit code) rather than an extension of this bugfix — recommend
  tracking separately rather than expanding this PR's scope.

https://claude.ai/code/session_01XWYzvHWU3EPgrZzU4T5JA5


aaronteoh:
## Review summary (Claude Code — implementation session)

Confident this is complete and correct, based on:

**Two independent expert code reviews** (separate agents, no context sharing with implementation):
- Review 1 (after initial fix): found the jsonl entry was missing `run_id`/`started_at`, and log files were written to the wrong directory (`scripts/{name}_{ts}/` instead of `scripts/runs/{run_id}/`, which is what the log-viewer route actually reads) — both fixed.
- Review 2 (after that fix, scoped to the new diff): found the failure-path error message written to history was weaker than what activity log/notifications showed, and a cosmetic redundant import — both fixed.

**Verification performed:**
- `pytest -x`: 4209 passed, 30 skipped, 0 failed (run after every commit)
- `ruff check` / `ruff format --check`: clean
- Manually traced `git diff origin/v1.0.7...HEAD` to confirm no other code depends on the old log directory layout before renaming it, and that the deferred issues below are genuinely pre-existing, not introduced by this PR

**Deferred, not fixed here (documented in PR description):**
1. Schema duplication across the two jsonl writers (`jobs.py` + `scripts_api.py`) — no shared type. Would need a shared module; out of scope for this bugfix.
2. Unlocked concurrent jsonl append — pre-existing pattern from the manual-run writer, this PR adds a second caller of it, not a new risk class.
3. Scheduled-run cancellation / pre-launch-exception paths still write no history entry — confirmed via git diff that this predates the PR and was never in scope; closing it is new functionality (no stdout/stderr/exit code exists for a pre-launch cancellation), not a fix.

None of the three deferred items are regressions introduced by this PR.

PR #429, opened 2026-08-26
fix: retry DuckDB lock conflicts in _load_with_lock

aaronteoh:
## Summary
- Add retry logic to `_load_with_lock()` for DuckDB file lock conflicts
- On local deployments, Metabase holds a continuous DuckDB connection; scheduled syncs
  intermittently fail with "Could not set lock" when Metabase is mid-query
  (`stop_metabase_for_writes()` is a no-op unless `DANGO_CLOUD_MODE=true`)
- Fix: retry `pipeline.load()` up to 5 times with 10s backoff when the caught
  `DestinationConnectionError` message contains "Could not set lock"
- Cloud deployments unaffected — Metabase is already stopped before `pipeline.load()` runs there
- All 3 call sites of `_load_with_lock` (`_run_dlt_native_source`, `_run_dlt_source`, and its
  partial-sync retry branch) are covered automatically since they share this one method — none
  of them were touched directly

## Verification
- `pytest -x`: 4207 pass, 30 skipped, 0 fail
- `pytest tests/unit/test_duckdb_lock_retry.py -x -v`: 3 pass
- `ruff check dango/`: pass
- `ruff format --check dango/`: pass

## Regression check
- Non-lock `DestinationConnectionError` (e.g. wrong credentials) raised immediately — tested, no retry
- Any other exception type is not caught by the retry loop at all — propagates immediately
- `DbtLock` is held for the full duration of retries, so other syncs wait rather than race
- Both `_run_dlt_source` and `_run_dlt_native_source` covered via the shared `_load_with_lock`

https://claude.ai/code/session_01TZvevjDU4GdUP7VNTGaeHe


aaronteoh:
## Code review update

Ran an independent code review on this PR and it caught a critical bug: the original retry code caught `DestinationConnectionError` directly, but dlt's `Pipeline.load()` always wraps step exceptions in `PipelineStepFailed` (confirmed by reading dlt's source — `Pipeline.load()` does `raise PipelineStepFailed(self, "load", err_load_id, l_ex, step_info) from l_ex` around the whole load step). That means the `except DestinationConnectionError` clause could never match in production — a real DuckDB lock conflict would raise on the very first attempt, and the retry loop was dead code. The unit tests passed only because they mocked `pipeline.load.side_effect` to raise `DestinationConnectionError` directly, which doesn't match dlt's real behavior.

Also flagged: the trailing `raise _last_lock_error` after the loop was unreachable — every loop iteration either returns or raises via the `else: raise` branch on the final attempt, so that line could never execute.

Fixed in the latest commit (d782b50):
- Now catches `PipelineStepFailed` instead. Verified live that `PipelineStepFailed`'s message embeds the wrapped exception's `str()`, so the `"Could not set lock"` substring match still works correctly against the wrapped error.
- Removed the unreachable trailing raise; replaced with an explicit `raise AssertionError("unreachable: ...")` for type-checker completeness.
- Updated the 3 unit tests to construct real `PipelineStepFailed`-wrapped errors (matching what dlt actually raises) instead of raising the inner exception directly.

Re-verified: `pytest -x` 4207 passed / 30 skipped, `pytest tests/unit/test_duckdb_lock_retry.py -x -v` 3 passed, `ruff check`/`ruff format --check` clean.

A third finding (DbtLock held during the retry sleeps, delaying other pending syncs by up to ~40s) is expected/accepted behavior per this task's original spec, not a bug — noted for visibility during merge review.

aaronteoh:
## Second code review pass (0ff4333)

Ran an independent review again on the fixed code (given how significant the first finding was, worth a second pair of eyes). It surfaced one more real bug and reiterated one already-accepted tradeoff:

**Fixed:** the retry check only matched the literal substring `"Could not set lock"`, but this exact file already has an established, tested keyword set for classifying DuckDB lock errors in `_analyze_error()` (BUG-195, `tests/unit/test_analyze_error.py`): `"is locked"`, `"could not set lock"`, `"write lock"`, `"file lock"`, `"already open"`, `"another process"` (case-insensitive). A real lock conflict phrased differently (e.g. `"database is locked"`, `"write lock held by another process"`) would have skipped the retry and failed on the first attempt, silently missing exactly the class of error this PR exists to handle for those variants.

Extracted a shared `_is_duckdb_lock_error()` predicate (plus the keyword tuple) used by both `_analyze_error()` and `_load_with_lock()`, so the two can't drift apart again. Added a parametrized test (`test_retries_on_other_lock_phrasings`) proving the retry now fires for all the BUG-195 phrasings, not just the one literal message from the incident traceback.

**Not changed — reiterates an already-accepted tradeoff:** `DbtLock` stays held for the full duration of the retry sleeps (up to ~40s across 4 retries). The review noted a theoretical edge case where a second sync waiting near its own `max_lock_wait` (default 300s) could now time out due to the longer hold. This is the same tradeoff the original task spec explicitly called out and accepted ("better than failing outright"), and the task's own constraints forbid changing the `DbtLock` acquire/release structure — releasing the lock mid-retry would also be unsafe, since `DbtLock` exists specifically to prevent two of Dango's own sync processes from racing for the same DuckDB file. Leaving as-is.

**Not changed — out of scope:** a suggestion to expose `_LOCK_MAX_RETRIES`/`_LOCK_RETRY_WAIT` as method parameters (like `max_lock_wait`) for tunability. This conflicts with the task's explicit instruction not to change `_load_with_lock`'s signature, and isn't needed for the incident this PR fixes — noted for visibility only.

Re-verified: `pytest -x` 4211 passed / 30 skipped (incl. the pre-existing `test_analyze_error.py` suite, unaffected by the extraction), `pytest tests/unit/test_duckdb_lock_retry.py -v` 7 passed, `ruff check`/`ruff format --check` clean.

aaronteoh:
## Coordinating chat review — APPROVED

**Critical improvement:** The session correctly caught that `pipeline.load()` wraps underlying exceptions in `PipelineStepFailed` — the prompt's `DestinationConnectionError` catch would have never triggered. The session verified actual dlt behavior and fixed the spec. Good catch.

**`_DUCKDB_LOCK_KEYWORDS` extraction:** Refactored the duplicate keyword list from `_analyze_error()` into a module-level constant shared with the retry logic. Both now stay in sync as DuckDB/dlt error wording changes. Better than just checking `"Could not set lock"` as the prompt said.

**`raise AssertionError("unreachable")` sentinel:** Cleaner than the prompt's `_last_lock_error: Exception | None` pattern (no `# type: ignore` needed, control flow is explicit).

**Tests:** More thorough than specified — parametrized over all 4 lock keyword phrasings, handles the `dango.utils.dbt_lock` module shadowing gotcha, asserts exact sleep call with `mock_sleep.assert_called_once_with(10)`.

**Minor note:** Logging uses `_logging.getLogger(__name__).warning(...)` (standard logging format args) instead of the structlog-style `logger.warning("event", key=val)` the prompt specified. Functionally fine — the retry attempt gets logged either way. Not worth blocking on.

4207 tests pass. PR #429 approved.

PR #430, opened 2026-08-26
fix: consolidate variables page into secrets page

aaronteoh:
## Summary
- Remove duplicate /settings/variables page (env vars are secrets, belong in one place)
- Improve env var masking: show partial value (last 4 chars, e.g. `****cdef`) instead of blanket `***`
- Add 301 redirect from /settings/variables → /settings/secrets for any bookmarks
- Remove variables.py router and variables.html template entirely

## Deviations from the session prompt (both verified against actual code, not assumed)

1. **Masking scheme**: the prompt's literal `_mask_value` snippet masks the first 2 chars
   (`ab****`), but the actual `variables.py` it claimed to "copy from" masks the **last 4
   chars** (`****cdef`). Confirmed with the requester — implemented the real variables.py
   behavior (last 4 chars), not the prompt's snippet.
2. **Missing test file**: `tests/unit/test_variables_route.py` imports
   `dango.web.routes.variables` directly and wasn't mentioned in the prompt. Deleted it
   alongside `routes/variables.py` — otherwise `pytest -x` fails on collection.
3. Also updated `tests/unit/test_web_secrets.py::test_list_with_env_vars`, which asserted
   the old blanket `***` masking, and strengthened it to actually exercise partial masking
   (fixture previously only used values ≤4 chars, which can't distinguish the two masking
   schemes).

## Verification
- `pytest -x` (full suite): 4181 passed, 30 skipped, 0 failed
- `pytest tests/unit/test_web_imports.py -x -v`: 16 passed
- `ruff format --check dango/`: 231 files already formatted
- Route registration confirmed via `TestClient` + `iter_route_contexts`: `/settings/secrets`
  and `/settings/variables` (redirect) both registered, `/api/variables/*` fully gone,
  `/api/secrets` unchanged
- `TestClient.get("/settings/variables")` → confirmed `301` to `/settings/secrets`

## Regression check
- `/api/secrets` response envelope unchanged (`env_vars` + `oauth_credentials`)
- OAuth section on secrets page unaffected — not touched
- Modal structure in secrets.html unaffected — only the masked-value `<td>` changed
- `grep -rn "api/variables" dango/web/templates/ dango/web/static/` → zero results
- `grep -n "settings/variables" dango/web/templates/base.html` → zero results
- Full repo sweep for `routes.variables|variables.html|settings/variables|routes/variables`
  → only remaining match is the intentional redirect route in `secrets.py`


aaronteoh:
## Independent code review

Ran an independent high-effort review against this PR's diff (`v1.0.7...fix/107-secrets-consolidation`). Verified each finding against the actual code before acting — two were real and fixed, two were dropped after verification.

**Fixed (commit 9bddd98):**
- `variables_redirect()` (the new `/settings/variables` → `/settings/secrets` route) had no permission dependency, unlike every other endpoint in `secrets.py` — whose module docstring states *"All endpoints require config.manage permission (admin-only)."* Not exploitable (the redirect target already enforces `config.manage`, and `AuthMiddleware` default-denies unauthenticated requests to this path anyway), but a logged-in non-admin could reach the handler itself. Added `Depends(require_permission("config.manage"))` to match the rest of the file.
- No test covered the new redirect. Added `TestVariablesRedirect::test_redirects_to_secrets_page`, asserting the 301 + `Location: /settings/secrets`.

**Verified and dropped:**
- *"Two inconsistent masking implementations"* — `mask_sensitive_config()` in `helpers.py` is unrelated; it's used only by `routes/sources.py` to recursively mask a data-source config dict, a different shape for a different endpoint. Not a duplicate of `_mask_value`.
- *"Legacy redirect should be a reusable mechanism, not a one-off"* — this PR retires exactly one page. Building a generic legacy-URL registry for a single redirect route is the premature abstraction this repo's `CLAUDE.md` explicitly warns against.

**Noted, no action taken:**
- The last-4-chars masking (`_mask_value`) already existed on the now-deleted `/settings/variables` page at the same `config.manage` permission level — this PR doesn't introduce new exposure, it consolidates existing, already-decided behavior onto the canonical page.

Full suite after the fix: 4182 passed, 0 failed. `ruff check`/`format` clean.

aaronteoh:
## Final review — approved

Confirming everything is complete and correct before merge.

**Scope check:** diff is exactly the variables→secrets consolidation described in the PR body — no unrelated changes.

**Verified:**
- `_mask_value`: `"****" + value[-4:]` for values >4 chars, else `"****"` — matches the real (pre-existing) `variables.py` logic, per your confirmation.
- `list_secrets()`: `for k, v in env_vars.items()` — correctly threads the value through to `_mask_value(v)`, no `KeyError`/iteration bug.
- `variables_redirect()`: now gated by `Depends(require_permission("config.manage"))`, consistent with every other endpoint in the file (fixed after my first review flagged it as missing).
- Redirect test asserts `301` + `Location: /settings/secrets` with `follow_redirects=False`.
- `routes/variables.py`, `templates/variables.html`, `tests/unit/test_variables_route.py` all deleted; repo-wide grep confirms no dangling references except the intentional `/settings/variables` redirect path string itself.
- Nav links removed from both desktop dropdown and mobile menu in `base.html`.
- `/api/secrets` response envelope unchanged; OAuth section and modal structure in `secrets.html` untouched.

**One thing I caught that wasn't in either review pass:** the compiled Tailwind CSS (`tailwind.min.css`) hadn't been rebuilt after the template edits, so CI's "CSS build (tailwind)" drift check (`make css-build` + `git diff --exit-code`) was failing on every commit in this PR. Fixed by running `make css-build` per `web/CLAUDE.md`'s documented workflow and committing the output.

**CI status:** all 8 checks green (Lint, mypy, Python 3.10/3.12/3.13, pip-audit, LLM checks, CSS build). Full local suite: 4182 passed, 0 failed.

Merging.

PR #438, opened 2026-08-31
1.0.8-B: Egress allowlist CI test

aaronteoh:
## Summary
- Add `docs/network-egress.yml` — enumerated, source-verified list of hosts the Dango
  stack contacts (telemetry hosts and env vars corrected against the actual installed
  dbt-core/dlt packages, not vendor docs — see corrections below)
- Add `egress` job to `.github/workflows/ci.yml` — live-detection test, not just a
  documentation-consistency check: patches Python-layer DNS resolution and runs a real
  `dango init` -> CSV source -> sync workflow inside it, failing on any undocumented host
- Marimo's "no telemetry" claim verified against installed marimo source in this
  session (previously asserted but unverified — `LAUNCH-READINESS.md` §A5 flagged this
  gap explicitly)
- Tests are class-wrapped (`TestNetworkEgressDoc`, `TestLiveEgressDetection`) with
  markers applied at the class level per `STANDARDS.md` §7; the live-detection test
  gets its own class since it needs an extra `egress` marker the doc-consistency tests
  don't
- Host comparisons are case/trailing-dot normalized; doc entries are read with `.get()`
  and a clear assertion message instead of raw `entry["host"]` indexing, so a future
  typo'd YAML key fails with a pointer to the problem, not a bare `KeyError`

### Corrections made against an earlier draft of this spec (verified via `dango/venv`, not vendor docs)
| Claim | Verified reality | Source |
|---|---|---|
| dbt telemetry host: `firehose.us-east-1.amazonaws.com` | `fishtownanalytics.sinter-collect.com` | `dbt/tracking.py:35` |
| dbt opt-out env var: `DBT_ENGINE_SEND_ANONYMOUS_USAGE_STATS` | `DBT_SEND_ANONYMOUS_USAGE_STATS` (no `ENGINE_`) | `dbt/cli/params.py:569` |
| dlt telemetry host: `dlthub.com` | `telemetry.scalevector.ai` | `dlt/common/configuration/specs/runtime_configuration.py:19` |

Also documents two real `github.com` download paths found by tracing the actual code
(not previously written down anywhere): the Metabase DuckDB driver download on
`dango init` (`cli/init.py:_setup_metabase`) and the spaCy PII model download inside
post-sync hooks (`governance/pii_detector.py`) — both redirect through
`objects.githubusercontent.com`, now also documented.

### dlt telemetry coverage — verified empirically, corrected from an earlier overclaim
An earlier version of this PR's test docstring stated dlt telemetry "IS covered by
this test" because dlt runs in-process. That's incomplete in two ways, both now fixed:
1. dlt dispatches telemetry fire-and-forget on an unjoined background thread
   (`dlt/common/runtime/anon_tracker.py`) — a queued send's DNS resolution could
   execute after the guard exits, racing past detection. Fixed by calling
   `dlt.common.runtime.telemetry.stop_telemetry()` before the guard exits, which
   blocks (`ThreadPoolExecutor.shutdown(wait=True)`) until any pending send completes.
2. More fundamentally: I checked at runtime, and for the CSV-source path this test
   actually drives, dlt's telemetry runtime never initializes at all —
   `is_telemetry_started()` and `_ANON_TRACKER_ENDPOINT` both stay unset through a
   full `run_sync()` call. `stop_telemetry()` is a safe no-op today; it's kept as
   defense in depth for a different source type or future dlt/dango version that does
   touch dlt's runtime-config lazy-init. The docstring now says this explicitly instead
   of implying dlt telemetry was observed firing.

## Verification
- `pytest -x` (no `-m` filter, matches local dev): 4202 pass, 30 skipped, 0 fail
- `pytest --tb=short -q -m "not egress"` (matches the CI `test` job exactly):
  4201 pass, 30 skipped, 1 deselected, 0 fail
- `pytest tests/unit/test_egress_allowlist.py -v`: 5 pass, 0 fail
- `ruff format --check dango/ tests/`: all files formatted
- `mypy tests/unit/test_egress_allowlist.py`: no issues
- The live-detection test (`test_dango_init_source_sync_only_contacts_allowlisted_hosts`)
  runs a real `dango init --skip-wizard` + CSV source + `run_sync(skip_dbt=True)` inside
  a `socket.getaddrinfo` monkeypatch that blocks any hostname outside
  `docs/network-egress.yml` — passed with zero undocumented hosts contacted
- Positive-control check (not part of the test suite, run manually to prove the
  detection mechanism actually works rather than trusting it): deliberately excluded
  `telemetry.scalevector.ai` from the allowed set and re-ran the same workflow —
  correctly caught and reported as blocked

## Regression check
- CI job name is `Egress allowlist`. **Corrected verification method**: an earlier
  push checked this only against `main`'s branch protection
  (`gh api repos/getdango/dango/branches/main/protection`). PR #438 targets `v1.0.8`,
  which has no classic branch protection at all (`gh api
  repos/getdango/dango/branches/v1.0.8/protection` → 404) — the actual gate is a
  separate ruleset, "Protect version branches" (id 12778188,
  `gh api repos/getdango/dango/rulesets/12778188`), whose `required_status_checks` are
  identical: `Lint (ruff)`, `Type check (mypy)`, `Test (Python 3.10)`,
  `Test (Python 3.12)`, no `Egress allowlist`. Conclusion unchanged, verification method
  now correct.
- Added as a job inside the existing single `ci.yml` (repo convention), not a new
  workflow file
- No subprocess-level network interception attempted (dbt subprocess egress is
  explicitly out of scope — a parent-process monkeypatch can't see a child process's
  own socket calls); dlt itself runs in-process and its DNS calls are visible to the
  guard when it makes any (see "dlt telemetry coverage" above for what that does and
  doesn't currently mean in practice)
- Full test suite passes unchanged with the CI-matching filter (4201/4201 with
  `-m "not egress"`, pre-existing 30 skips unaffected)
- Local reproduction note: the post-sync PII-scan hook isn't gated by `skip_dbt`
  (`post_sync.py:381`, unconditional) — run `python -m spacy download en_core_web_sm`
  once before running the `egress`-marked test locally (matches the CI `egress` job's
  pre-install step), or it falls through to a live model download instead of a clean
  skip

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Er5nVqaH2DsnhCh58NTLXU

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,13 @@ jobs:
       - name: pytest
         env:
           DANGO_LOG_LEVEL: ERROR
-        run: pytest --tb=short -q
+        # -m "not egress": the egress-marked test needs the spaCy PII model
+        # pre-installed to stay hermetic (see the egress job's dedicated
+        # pre-install step below) — this matrix doesn't have that step, so
+        # running it here would add a live GitHub Releases download
+        # dependency to these required checks for no benefit; the egress
+        # job already covers it.
+        run: pytest --tb=short -q -m "not egress"
 
   security:
     name: Security (pip-audit)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,27 @@ jobs:
       - name: pip-audit
         run: pip-audit --ignore-vuln CVE-2026-32274 --ignore-vuln PYSEC-2026-2120 --ignore-vuln CVE-2026-71491 --ignore-vuln CVE-2026-59894 --ignore-vuln CVE-2026-59893 --ignore-vuln CVE-2026-54284
 
+  egress:
+    name: Egress allowlist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+      - run: pip install -c constraints.txt -e ".[dev]"
+      - name: Pre-install spaCy PII model
+        # Real download, done here (unguarded) so the guarded test below is
+        # hermetic — spacy.load() succeeds without any network call once this
+        # is installed. Without this step, the live-detection test's PII scan
+        # hook falls through to governance/pii_detector.py's urllib fallback
+        # and either does a slow live download or gets correctly blocked as
+        # an "unexpected host" (it isn't unexpected, it's just not pre-empted).
+        run: python -m spacy download en_core_web_sm
+      - name: Egress allowlist test
+        run: pytest tests/unit/test_egress_allowlist.py -v
+
   llm-checks:
     name: LLM checks
     runs-on: ubuntu-latest
--- a/docs/network-egress.yml
+++ b/docs/network-egress.yml
@@ -1,7 +1,10 @@
 # Dango network egress — all outbound connections made by the stack.
-# Telemetry: controllable via `dango telemetry off --all`
+# Telemetry: each provider has its own opt-out today (see each entry's
+# `control` field). A unified `dango telemetry off --all` command is planned
+# (Session D, not yet merged as of this session) — do not assume it exists.
 # Functional: required for operation, disclosed but not toggleable
-# Verified against installed packages (dbt-core, dlt, marimo source) 2026-08-31.
+# Verified against installed packages (dbt-core 1.11.14, dlt 1.28.1, marimo)
+# 2026-09-01 — re-verified after the dbt-core 1.10->1.11 bump landed on v1.0.8.
 # Metabase's own telemetry endpoint is NOT independently source-verified in this
 # session (Metabase is a Java process, not inspectable from this Python venv) —
 # confirm against Metabase's own telemetry docs before relying on it publicly.
@@ -9,7 +12,13 @@
 telemetry:
   - host: telemetry.getdango.dev
     provider: dango
-    control: "dango telemetry off --provider dango"
+    control: >
+      Today: decline the `dango init` consent prompt (persists "no"), or set
+      DO_NOT_TRACK=1 / DANGO_TELEMETRY=0 / `telemetry: false` in
+      ~/.dango/config.yml (dango/telemetry.py:is_telemetry_enabled) — all
+      three suppress the prompt entirely, no ping is ever sent. A unified
+      `dango telemetry off --provider dango` command is Session D's job
+      (dango/cli/commands/ has no `telemetry.py` yet as of this session).
     sends: "anonymous UUID, version, OS, source type names, weekly heartbeat"
   - host: fishtownanalytics.sinter-collect.com
     provider: dbt-core
@@ -21,7 +30,11 @@ telemetry:
     sends: "command names, hashed pipeline/dataset names, execution times, OS"
   - host: metabase.com
     provider: metabase
-    control: "dango telemetry off --provider metabase (sets allow_tracking=false via API)"
+    control: >
+      Planned: `dango telemetry off --provider metabase` (sets
+      allow_tracking=false via the Metabase admin API) is Session D's job,
+      not yet implemented as of this session — today this requires toggling
+      the setting manually in the Metabase admin UI.
     sends: "anonymous usage statistics"
 
 functional:
--- a/docs/network-egress.yml
+++ b/docs/network-egress.yml
@@ -1,0 +1,58 @@
+# Dango network egress — all outbound connections made by the stack.
+# Telemetry: controllable via `dango telemetry off --all`
+# Functional: required for operation, disclosed but not toggleable
+# Verified against installed packages (dbt-core, dlt, marimo source) 2026-08-31.
+# Metabase's own telemetry endpoint is NOT independently source-verified in this
+# session (Metabase is a Java process, not inspectable from this Python venv) —
+# confirm against Metabase's own telemetry docs before relying on it publicly.
+
+telemetry:
+  - host: telemetry.getdango.dev
+    provider: dango
+    control: "dango telemetry off --provider dango"
+    sends: "anonymous UUID, version, OS, source type names, weekly heartbeat"
+  - host: fishtownanalytics.sinter-collect.com
+    provider: dbt-core
+    control: "DBT_SEND_ANONYMOUS_USAGE_STATS=false or DO_NOT_TRACK=1"
+    sends: "OS, Python version, invocation success/duration, hashed model content (Snowplow)"
+  - host: telemetry.scalevector.ai
+    provider: dlt
+    control: "dlthub_telemetry=false in .dlt/config.toml, or dlt --disable-telemetry"
+    sends: "command names, hashed pipeline/dataset names, execution times, OS"
+  - host: metabase.com
+    provider: metabase
+    control: "dango telemetry off --provider metabase (sets allow_tracking=false via API)"
+    sends: "anonymous usage statistics"
+
+functional:
+  - host: extensions.duckdb.org
+    reason: "DuckDB autoloads extensions on demand"
+    controllable: false
+  - host: hub.getdbt.com
+    reason: "dbt package installation (dbt deps)"
+    controllable: false
+  - host: pypi.org
+    reason: "dango upgrade command"
+    controllable: false
+  - host: registry-1.docker.io
+    reason: "Metabase Docker image pull on first start"
+    controllable: false
+  - host: github.com
+    reason: >
+      Two separate download paths: (1) spaCy PII model wheel on first Presidio use
+      (governance/pii_detector.py, github.com/explosion/spacy-models), (2) Metabase
+      DuckDB driver on `dango init` if not cached (utils/driver.py,
+      github.com/motherduckdb/metabase_duckdb_driver)
+    controllable: false
+  - host: objects.githubusercontent.com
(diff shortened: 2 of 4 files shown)


aaronteoh:
Ran an independent code review of this PR. It found three real issues, all now fixed in the latest commit:

1. **Real dbt telemetry ping fired on every test run.** `init_project()`'s unconditional `dbt docs generate` step (`cli/init.py:100`, treated as CRITICAL — rolls back init on failure) inherits the parent process env with no override, so it was sending a real ping to `fishtownanalytics.sinter-collect.com` every time the live-detection test ran, invisible to the test's own DNS-guard since dbt runs as a subprocess. Fixed by setting `DBT_SEND_ANONYMOUS_USAGE_STATS`/`DO_NOT_TRACK` via `monkeypatch.setenv` before the guarded block — an env var, not subprocess interception, so it stays in scope.

2. **The pre-existing `test` job matrix (3.10/3.12/3.13) was also collecting the new live-detection test**, without the `egress` job's spaCy pre-install step. That meant the test would fall through to a live GitHub Releases model download three times per CI run — adding a real network dependency to the *required* `Test (Python 3.10)`/`Test (Python 3.12)` checks that wasn't there before. Fixed with a new `egress` marker (registered in `pyproject.toml`) and `-m "not egress"` on the `test` job; the dedicated `egress` job still runs the full file, hermetically.

3. **`docs/network-egress.yml` was missing `files.pythonhosted.org`** for the `dango upgrade` command — `pypi.org` only resolves package metadata, the actual wheel download goes through PyPI's separate CDN host. Added, matching the same pattern already used for the GitHub Releases CDN entries.

All fixes verified: `pytest tests/unit/test_egress_allowlist.py -v` (5 pass), full suite with `-m "not egress"` (4201 pass, 30 skipped, matching the `test` job), `ruff check`/`format --check`/`mypy` clean.

aaronteoh:
Fixed all three points from the coordinating-chat review:

1. **Wording** — module docstring no longer calls itself a "CI gate"; now states plainly it's informational, not merge-blocking (matches the actual branch protection: confirmed neither \`main\` nor the \`v1.0.8\`-covering ruleset lists \`Egress allowlist\` as required).
2. **Marker convention** (real violation, fixed) — wrapped all 5 tests into two classes (\`TestNetworkEgressDoc\`, \`TestLiveEgressDetection\`) with markers applied at the class level per STANDARDS.md §7. Split into two classes rather than one since the live-detection test needs the extra \`egress\` marker the doc-consistency tests don't. Verified \`-m "not egress"\`/\`-m "egress"\` still select exactly the same tests as before via \`--collect-only\`.
3. **Host normalization** — added lowercase + trailing-dot stripping in \`_AllowlistGuard\` and \`_documented_hosts\`, so case/trailing-dot variants of a documented host don't false-positive as undocumented.

Also ran a second independent review scoped to the previous fix commit (886ad1e) before this round — it came back with zero findings, verifying (not just trusting) the dbt env var behavior, the dlt/dbt source claims, and the marker exclusion.

Verified: `pytest tests/unit/test_egress_allowlist.py -v` (5 pass), full suite with `-m "not egress"` (4201 pass, 30 skipped, matching what the `test` job runs), `ruff check`/`format --check`/`mypy` clean.

aaronteoh:
Fixed all items from the second review round. Findings #1–5 addressed (#6 and #7 were flagged as not-fixable/optional style — agreed, no action taken on those):

1. **dlt telemetry race — real, went further than the finding.** Verified the reviewer's claim against dlt's source (`anon_tracker.py`, `managed_thread_pool.py`) — confirmed the fire-and-forget background thread is real. Fixed with `dlt.common.runtime.telemetry.stop_telemetry()` before the guard exits, forcing a synchronous flush. Then checked something the finding didn't raise: **does dlt telemetry actually fire at all for the CSV-source path this test drives?** Empirically, no — `is_telemetry_started()`/`_ANON_TRACKER_ENDPOINT` stay unset through a full `run_sync()` call. So the original "dlt IS covered" claim was overclaiming in a more basic way than "racy" — there's currently no live event to race against. `stop_telemetry()` is kept as defense in depth; docstring now says this precisely rather than implying observed behavior. Verified the detection mechanism itself works with a manual positive control (excluded dlt's host from the allowlist, confirmed it gets caught).
2. **PR body verification method corrected** — it cited `main`'s branch protection; `v1.0.8` (the actual target) has none (404), the real gate is the "Protect version branches" ruleset. Same conclusion, now cited correctly. Full PR body re-audited and updated, not just a new comment.
3. **Local dev friction** — noted in the docstring: pre-install the spaCy model before running the `egress`-marked test locally, or it falls through to a live download.
4. **Host normalization on the failure-message side** — `blocked_hosts` now stores the normalized form, matching what someone would paste into `docs/network-egress.yml`.
5. **KeyError robustness** — `_documented_hosts()` and the provider test now use `.get()` with a clear assertion message instead of raw dict indexing.

Verified: `pytest tests/unit/test_egress_allowlist.py -v` (5 pass), full suite with `-m "not egress"` (4201 pass, matching the `test` job), `ruff`/`mypy` clean, plus a manual positive-control run proving the detection mechanism catches an excluded host.

aaronteoh:
Before merging: two PRs landed on \`v1.0.8\` since this branch was cut — #439 (dbt-core 1.10→1.11.14, dbt-duckdb 1.11, Metabase v0.62.18) and #440 (Dango's own opt-in telemetry: \`dango/telemetry.py\` + a consent prompt in \`cli/init.py\`). Merged \`v1.0.8\` into this branch and re-verified rather than assuming compatibility:

- Re-checked dbt's telemetry host/env var directly against the newly installed dbt-core 1.11.14 — both unchanged, still accurate. dlt unchanged.
- \`cli/init.py\`'s new consent prompt is called unconditionally in \`initialize()\` (not gated by \`skip_wizard\`) and could call \`click.prompt()\` interactively or send a real ping. Ran the egress test against the merged code to check empirically rather than assume — it passed cleanly with no hang, because \`DO_NOT_TRACK=1\` (already set in an earlier fix round, for dbt) is also honored by Dango's own \`is_telemetry_enabled()\`. Confirmed by reading that function, not just by the test passing. Documented explicitly: real CI is independently protected by \`is_ci()\`, but a contributor running this test locally on a fresh machine would hit a live prompt without \`DO_NOT_TRACK\` also covering it.
- \`docs/network-egress.yml\`'s dango/metabase entries described a \`dango telemetry off --provider X\` command as if it existed today. It doesn't — no \`cli/commands/telemetry.py\`, not registered in \`main.py\`. That's Session D, not yet merged. Corrected both \`control\` fields to describe what actually works today (env vars, declining the init prompt, \`~/.dango/config.yml\`), with an explicit note that the unified command is planned, not current.

Full suite re-run against the merged state: 4275 passed, 30 skipped, 1 deselected (up from 4201 pre-merge, reflecting new tests from #439/#440), 0 failed. ruff/mypy clean. All 5 egress tests still pass.

Re-running CI now on the merge commit before this is ready to merge.

aaronteoh:
## Final review — approved for merge

This PR went through four review rounds (2 by me, 2 by the coordinating chat) plus a re-verification against \`v1.0.8\`'s current state after two other PRs landed underneath this branch. Summary of what's actually being merged:

**Core design:** live detection, not documentation-only. \`_AllowlistGuard\` patches \`socket.getaddrinfo\` (the correct interception point — it still has the hostname string before DNS resolution, unlike patching \`connect()\`) and runs a real \`dango init\` → CSV source → \`sync\` workflow inside it. Records blocked hosts on \`self\` rather than relying on the raised exception reaching the test, since \`dispatch_post_sync_hooks\` swallows per-hook exceptions — verified this matters by tracing the actual swallowing code.

**What's been independently verified, not assumed, over the course of this PR:**
- dbt-core's telemetry host and env var name, and dlt's telemetry field name — checked against the actually-installed packages both before and after the mid-review dbt-core 1.10→1.11.14 bump landed on \`v1.0.8\`. All still accurate.
- Marimo's "no telemetry" claim — read marimo's installed source directly (this was previously asserted, never verified, per \`LAUNCH-READINESS.md\` §A5's own flag).
- dlt's fire-and-forget telemetry thread — traced the actual \`ManagedThreadPool\`/\`ThreadPoolExecutor\` mechanism, fixed with a synchronous flush (\`stop_telemetry()\`), then went further and confirmed empirically that telemetry doesn't even initialize for the code path this test drives — the docstring says this precisely instead of overclaiming coverage.
- Session C's new telemetry consent prompt (\`cli/init.py\`, added after this branch was cut) — confirmed by running the test against the merged code that \`DO_NOT_TRACK=1\` (already set for an unrelated reason) also correctly suppresses it, with no hang; documented why that matters for local reproduction specifically, since real CI is independently protected by \`is_ci()\`.
- \`docs/network-egress.yml\`'s \`dango telemetry off --provider X\` control fields — checked that this command doesn't exist yet (Session D, not merged) and corrected both entries to describe what actually works today.

**Scope held throughout:** dbt/dlt subprocess-level interception stayed explicitly out of scope (would need elevated CI permissions); the \`egress\` job stays informational (confirmed against the actual \`main\` branch protection and the \`v1.0.8\`-governing "Protect version branches" ruleset, not assumed).

**Verification on the current commit (\`efbc912\`):** all 9 CI checks pass, including \`Egress allowlist\` itself. Full suite 4275 passed / 30 skipped / 0 failed against the merged \`v1.0.8\` state. \`ruff\`/\`mypy\` clean.

Approved. Merging.

PR #440, opened 2026-08-31
1.0.8-C: Dango opt-in telemetry

aaronteoh:
## Summary
- Adds `dango/telemetry.py`: a single opt-in "install ping" fired once at `dango init`, gated behind an explicit consent prompt. No heartbeat/scheduler hook — out of scope for this session.
- `ProjectInitializer._prompt_telemetry_consent()` (`dango/cli/init.py`) wires the prompt in right after the success message prints. Call-site guard is `if not skip_wizard:`; the method itself additionally no-ops (without ever prompting) when `is_ci()`, `has_recorded_consent()`, or `not is_telemetry_enabled()` — i.e. CI detection, a prior stored answer, or any opt-out (`DO_NOT_TRACK`, `DANGO_TELEMETRY`, `telemetry: false` in `~/.dango/config.yml`) all skip silently before ever asking.
- The consent prompt itself is asked via a dedicated `_ask_telemetry_consent()` helper (module-level in `init.py`), not `safe_confirm()` — it calls `click.prompt()` directly and returns `None`, rather than substituting a default, when no real answer could be obtained at all (genuine EOF, or a closed stdin). This distinction matters because the answer gets persisted permanently: a real "no" is persisted, but an unseen non-answer never is. It also means a real answer piped into stdin (e.g. `echo yes | dango init`) is read and honored normally, since it doesn't pre-check `isatty()`.
- Identity is machine-level: anonymous UUID in `~/.dango/telemetry.json`, never project-scoped.
- Opt-out (any one disables): `DO_NOT_TRACK`/`DANGO_TELEMETRY` set to any spelling dbt-core itself accepts (`1/t/true/y/yes` and `0/f/false/n/no`, case-insensitive), `telemetry: false` in `~/.dango/config.yml`, answering "no" at the prompt, or running in CI (checks `CI`, `GITHUB_ACTIONS`, `GITLAB_CI`, `JENKINS_URL`, `BUILDKITE`, `CIRCLECI`, `TRAVIS`, `CODEBUILD_BUILD_ID`).
- Payload: anonymous UUID, event name, Dango version, OS, Python version, and source *type* names (enabled sources only) — never source names, credentials, row counts, schema, or query text.
- `ping()` runs the network call on a daemon thread joined with a `_TIMEOUT_SECONDS` (2s) timeout — bounded wait, not a fully synchronous call, but also not left un-joined (see below for why).
- No new pip dependency — stdlib `urllib.request` only (PyYAML was already a dependency).

### Issues found and fixed across three independent review passes
This PR went through three rounds of independent review (not self-review) between initial implementation and merge-readiness. Each caught something real:

**Round 1** — the original task spec's hook point (`ProjectInitializer.setup_project()` in `commands/project.py`) doesn't exist; wired into `ProjectInitializer.initialize()` instead. `config.sources` corrected to `config.sources.sources`. **Critical bug caught via live-endpoint testing:** Cloudflare's bot-fight mode 403s requests with no `User-Agent` header (urllib sends none by default) — every real ping would have silently failed forever. Fixed by adding `User-Agent: dango-cli/<version>`.

**Round 2** — (a) a non-interactive session's unseen `safe_confirm()` EOF-fallback answer was being persisted as a real "no", permanently locking a machine into opt-out without ever asking; (b) `DO_NOT_TRACK`/`DANGO_TELEMETRY` only matched `"1"`/`"0"` exactly, diverging from dbt-core's own accepted spellings; (c) `is_ci()` misread `CI=false` as CI; (d) `ping()` blocked synchronously; (e) disabled sources leaked into the payload; (f) a `not failures` guard was dead code (removed). A follow-up targeted review then caught that the TTY-check fix for (a) introduced its own unguarded crash when `sys.stdin` is `None` (closed fd) — fixed with a broader exception guard.

**Round 3** — the isatty()-based fix from round 2(a) turned out to be *too* broad: it silently dropped **any** non-TTY stdin, including a pipe with a real answer in it (`echo yes | dango init` got ignored entirely). Fixed by replacing the isatty() pre-check with the `_ask_telemetry_consent()` helper described above. Also found that the un-joined daemon thread from round 2(d)'s fix essentially never got to run at all before a real (undelayed) `dango init` process exit — reproduced directly: 3/3 clean runs wrote zero bytes to `~/.dango/telemetry.json`. Fixed by joining the thread with a timeout. Also removed a dead `is_ci` payload field that could never be anything but `False`.

Every fix above was verified by direct reproduction against the real code (not just trusting the review's claim) both before and after the change — see PR comments for the specific repro commands and outputs.

## Verification
- `pytest -x`: 4271 passed, 30 skipped, 0 failed (full suite — no regressions across all rounds of fixes)
- `pytest tests/unit/test_telemetry.py tests/unit/test_cli_init_telemetry.py -v`: 84/84 passed
- `ruff check` / `ruff format --check` / `mypy dango/telemetry.py dango/cli/init.py`: all clean
- `scripts/validate_headers.py` / `scripts/check_docstrings.py`: all clean
- **Live API verification (current state, re-verified after every round of fixes):**
  - `ping()` called for real (unmocked) against the live endpoint — confirmed `200 {"ok":true}`, `~/.dango/telemetry.json` correctly populated with `{"enabled": true, "uuid": "..."}`
  - A real `"yes\n"` piped into `_prompt_telemetry_consent()` is read and honored end-to-end: prompt text displayed, consent persisted, real ping fires
  - Empty/closed stdin (`< /dev/null`, and `sys.stdin` patched to `None` simulating a closed fd) returns `None` from `_ask_telemetry_consent()` — no prompt, nothing persisted, no crash
  - 3 consecutive real `dango init`-style invocations with **no artificial delay** confirmed the ping's UUID write (and, via an `urlopen` spy, the actual `200` response) now completes before process exit
  - `DO_NOT_TRACK`/`DANGO_TELEMETRY` opt-out confirmed for multiple accepted spellings via unit tests

## Regression check
- `dango init` project creation is unaffected by the telemetry prompt — it fires only after `_print_success_message()`, downstream of all project-creation steps.
- No existing test previously invoked `dango init` end-to-end, so nothing regressed from adding the prompt. The `skip_wizard` gate is covered by `test_skip_wizard_never_prompts`, which runs the real `initialize(skip_wizard=True)` path with only the heavy/unrelated steps (`_setup_metabase`, `_generate_dbt_docs`, `_setup_auth`) mocked.
- Docs kept in sync across all fix rounds: `dango/CLAUDE.md`, `dango/cli/CLAUDE.md` files tables, and `docs/file-exemptions.yml`'s line count for `cli/init.py` (1496 → 1585, with the reasons for each increment documented).

### Known, accepted follow-ups (not fixed in this PR — low severity, flagged by review)
- `_get_or_create_uuid()`/`set_telemetry_enabled()` do an unlocked read-modify-write of `~/.dango/telemetry.json`. Accepted risk: this only runs once per `dango init`, and the only realistic race is two `dango init` invocations landing on the same machine at literally the same instant, which would at worst regenerate a duplicate anonymous UUID — cosmetic, not a correctness bug.
- `_ask_telemetry_consent()` reimplements `safe_confirm()`'s yes/no parsing loop rather than sharing it, since it needs a different return contract (`bool | None` vs `bool`) and broader exception handling. A shared refactor is possible but touches ~25 other `safe_confirm()` call sites for marginal benefit — left as a future cleanup, not blocking.

https://claude.ai/code/session_011UBhTbqnQRga12Q3apJpKw

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -342,7 +342,7 @@ Full exemption registry: [`docs/file-exemptions.yml`](docs/file-exemptions.yml)
 | `ingestion/sources/registry.py` | 2340 | — (metadata-only) |
 | `cli/source_wizard.py` | 2525 | — |
 | `visualization/metabase.py` | 1251 | — |
-| `cli/init.py` | 1559 | — |
+| `cli/init.py` | 1585 | — |
 | `visualization/dashboard_manager.py` | 1117 | — |
 | `cli/commands/platform.py` | 1136 | — (extracted from main.py by TASK-005) |
 | `web/routes/auth.py` | 902 | — (split evaluated in DOC-025: exempt, security-critical) |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -342,7 +342,7 @@ Full exemption registry: [`docs/file-exemptions.yml`](docs/file-exemptions.yml)
 | `ingestion/sources/registry.py` | 2340 | — (metadata-only) |
 | `cli/source_wizard.py` | 2525 | — |
 | `visualization/metabase.py` | 1251 | — |
-| `cli/init.py` | 1549 | — |
+| `cli/init.py` | 1559 | — |
 | `visualization/dashboard_manager.py` | 1117 | — |
 | `cli/commands/platform.py` | 1136 | — (extracted from main.py by TASK-005) |
 | `web/routes/auth.py` | 902 | — (split evaluated in DOC-025: exempt, security-critical) |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -342,7 +342,7 @@ Full exemption registry: [`docs/file-exemptions.yml`](docs/file-exemptions.yml)
 | `ingestion/sources/registry.py` | 2340 | — (metadata-only) |
 | `cli/source_wizard.py` | 2525 | — |
 | `visualization/metabase.py` | 1251 | — |
-| `cli/init.py` | 1538 | — |
+| `cli/init.py` | 1549 | — |
 | `visualization/dashboard_manager.py` | 1117 | — |
 | `cli/commands/platform.py` | 1136 | — (extracted from main.py by TASK-005) |
 | `web/routes/auth.py` | 902 | — (split evaluated in DOC-025: exempt, security-critical) |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -342,7 +342,7 @@ Full exemption registry: [`docs/file-exemptions.yml`](docs/file-exemptions.yml)
 | `ingestion/sources/registry.py` | 2340 | — (metadata-only) |
 | `cli/source_wizard.py` | 2525 | — |
 | `visualization/metabase.py` | 1251 | — |
-| `cli/init.py` | 1496 | — |
+| `cli/init.py` | 1538 | — |
 | `visualization/dashboard_manager.py` | 1117 | — |
 | `cli/commands/platform.py` | 1136 | — (extracted from main.py by TASK-005) |
 | `web/routes/auth.py` | 902 | — (split evaluated in DOC-025: exempt, security-critical) |
--- a/dango/CLAUDE.md
+++ b/dango/CLAUDE.md
@@ -10,6 +10,7 @@ Root Python package for Dango — contains all module subpackages (cli, config, 
 |------|---------|----------------------|
 | `__init__.py` | Package metadata: version, author, license | `__version__`, `__author__`, `__license__` |
 | `logging.py` | Structured logging (structlog + stdlib integration) | `configure_logging()`, `get_logger()`, `bind_contextvars`, `clear_contextvars`, `unbind_contextvars` |
+| `telemetry.py` | Opt-in anonymous install telemetry (machine-level UUID in `~/.dango/`) | `is_ci()`, `is_telemetry_enabled()`, `has_recorded_consent()`, `set_telemetry_enabled()`, `ping()` |
 
 ## Common Tasks
 
@@ -30,6 +31,7 @@ Root Python package for Dango — contains all module subpackages (cli, config, 
 - `dango.logging` — `get_logger()` used by `web/routes/`, `web/middleware/`, `web/app.py`, `platform/scheduling/`, `platform/notifications/`, `oauth/web_flow.py`, `utils/log_rotation.py`, `utils/post_sync.py`, `cli/commands/schedule.py`, `cli/commands/remote_env.py`, `governance/`, `analysis/`; `configure_logging()` called by entry points. Note: `notebooks/` uses stdlib `logging` directly, not `dango.logging`.
 - `pyproject.toml` defines `getdango` as the installable package
 - Entry point: `dango.cli.main:cli` (configured in `pyproject.toml`)
+- `dango.telemetry` — `cli/init.py` (`ProjectInitializer._prompt_telemetry_consent()`) is the only caller; stdlib `urllib.request`/`json`/`platform`/`uuid` only, plus lazy `yaml` for the opt-out config file. No new pip dependency.
 
 ## Testing
 
@@ -37,6 +39,7 @@ Root Python package for Dango — contains all module subpackages (cli, config, 
 - **Unit only:** `pytest -m unit`
 - **Integration only:** `pytest -m integration`
 - **Logging tests:** `pytest tests/unit/test_logging.py`
+- **Telemetry tests:** `pytest tests/unit/test_telemetry.py tests/unit/test_cli_init_telemetry.py`
 
 ## Don't Modify
 
--- a/dango/cli/CLAUDE.md
+++ b/dango/cli/CLAUDE.md
@@ -48,7 +48,7 @@ Click-based command-line interface for all Dango operations — project init, so
 | `commands/dev.py` (~429 lines) | `dev` group (default run + clean) — branch-based dbt development | `dev` (group), `dev_clean()` |
 | `commands/web.py` (69 lines) | `web` dev server command | `web()` |
 | **Wizards** | | |
-| `init.py` (1549 lines) | Project initialization wizard, incl. first-run telemetry consent prompt | `ProjectInitializer` |
+| `init.py` (1559 lines) | Project initialization wizard, incl. first-run telemetry consent prompt | `ProjectInitializer` |
 | `wizard.py` (307 lines) | Interactive setup wizards | `ProjectWizard` |
 | `source_wizard.py` (2525 lines) | Source configuration wizard | `add_source()` |
 | `model_wizard.py` (547 lines) | dbt model creation wizard | `add_model()` |
--- a/dango/cli/CLAUDE.md
+++ b/dango/cli/CLAUDE.md
@@ -48,7 +48,7 @@ Click-based command-line interface for all Dango operations — project init, so
 | `commands/dev.py` (~429 lines) | `dev` group (default run + clean) — branch-based dbt development | `dev` (group), `dev_clean()` |
 | `commands/web.py` (69 lines) | `web` dev server command | `web()` |
 | **Wizards** | | |
-| `init.py` (1559 lines) | Project initialization wizard, incl. first-run telemetry consent prompt | `ProjectInitializer` |
+| `init.py` (1585 lines) | Project initialization wizard, incl. first-run telemetry consent prompt | `ProjectInitializer` |
 | `wizard.py` (307 lines) | Interactive setup wizards | `ProjectWizard` |
 | `source_wizard.py` (2525 lines) | Source configuration wizard | `add_source()` |
 | `model_wizard.py` (547 lines) | dbt model creation wizard | `add_model()` |
--- a/dango/cli/CLAUDE.md
+++ b/dango/cli/CLAUDE.md
@@ -48,7 +48,7 @@ Click-based command-line interface for all Dango operations — project init, so
 | `commands/dev.py` (~429 lines) | `dev` group (default run + clean) — branch-based dbt development | `dev` (group), `dev_clean()` |
 | `commands/web.py` (69 lines) | `web` dev server command | `web()` |
 | **Wizards** | | |
-| `init.py` (1496 lines) | Project initialization wizard | `ProjectInitializer` |
+| `init.py` (1538 lines) | Project initialization wizard, incl. first-run telemetry consent prompt | `ProjectInitializer` |
 | `wizard.py` (307 lines) | Interactive setup wizards | `ProjectWizard` |
 | `source_wizard.py` (2525 lines) | Source configuration wizard | `add_source()` |
 | `model_wizard.py` (547 lines) | dbt model creation wizard | `add_model()` |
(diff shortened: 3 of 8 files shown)


aaronteoh:
## Independent code review + fixes

Ran an independent review of this PR (separate pass, not the original author's self-check) and found 6 issues, all verified against the real code/behavior before fixing — 2 high, 2 medium, 2 low. Fixed in the two follow-up commits above.

**High**
1. **Silent permanent opt-out lock-in.** `safe_confirm()`'s non-interactive fallback (used when stdin isn't a TTY) returned its default `False` *without ever showing the prompt*, but that unseen fallback was being persisted via `set_telemetry_enabled(False)` as if it were a real answer. Any non-interactive `dango init` not caught by CI-env-var detection (devcontainer setup, SSH without a pty, provisioning scripts) would permanently lock that machine into opt-out — the human is never asked again on any future project. **Fix:** added an explicit `sys.stdin.isatty()` check before calling `safe_confirm()`; if not a TTY, skip without persisting anything, so the real user still gets asked next time they run `dango init` interactively. Verified manually: a non-TTY run now creates no `~/.dango/` state at all (previously it wrote `enabled: false`).

2. **`DO_NOT_TRACK` didn't match its own cited precedent.** Verified against dbt-core's actual source in the venv (`dbt/cli/flags.py`) — it accepts `1/t/true/y/yes` case-insensitively. This module only matched the exact string `"1"`, so `DO_NOT_TRACK=true` (arguably the more natural spelling) silently failed to opt out, despite `LAUNCH-READINESS.md` explicitly citing "the cross-tool standard dbt already respects" as the reason to honor it. **Fix:** matched dbt's exact accepted value sets for both `DO_NOT_TRACK` (truthy) and `DANGO_TELEMETRY` (falsy).

**Medium**
3. **`ping()` blocked synchronously**, contradicting the module's own "fire-and-forget, never blocks" design intent. **Fix:** the network call now runs on an un-joined daemon thread. Verified manually: `ping()` now returns in <1ms while the backgrounded request still completes with `200` against the live endpoint (confirmed via an `urlopen` spy).

4. **Disabled sources leaked into the install ping.** `source_types` was built from `config.sources.sources` (all sources) instead of the existing `SourcesConfig.get_enabled_sources()` filter, skewing the adoption signal. Fixed.

**Low**
5. `is_ci()` treated any non-empty env var as CI, so an explicit `CI=false` would be misread as CI=true (fail-safe direction, narrow case). Fixed with the same truthy/falsy value-matching approach as #2.
6. The `if not failures` half of the telemetry-prompt guard was dead code — confirmed via grep that `failures` is never appended to anywhere in `initialize()`. Removed, and the adjacent comment corrected.

All fixes covered by new/updated tests (68 tests now in the two telemetry test files, up from 35); full suite at 4265 passed, 0 failed, no regressions. Manual live-endpoint re-verification done for both the TTY fix and the threading fix.

aaronteoh:
## Targeted re-review of the fix commit

Ran a second, narrower review scoped specifically to the previous fix commit (\`5e14df4\`) — checking whether the 6 fixes were correct/complete and whether they introduced any new regressions, rather than re-covering the whole PR.

**All 6 original fixes verified correct and complete** — independently re-confirmed against the current code (including cross-checking the DO_NOT_TRACK value set against dbt-core's actual source again), and the threading tests genuinely exercise non-blocking behavior with a real unmocked thread (not just mocks that would pass even if the fix were wrong).

**One new issue found and fixed** (commit \`0044050\`):

The new \`sys.stdin.isatty()\` check (added to fix the TTY lock-in issue) was itself unguarded. When fd 0 is closed entirely — not just redirected, e.g. a daemonized invocation — CPython sets \`sys.stdin\` to \`None\`, and \`None.isatty()\` raises \`AttributeError\`. That propagated straight up through \`initialize()\` right after it had already printed the "Success" panel, crashing \`dango init\` with an unhandled traceback despite the project being fully and correctly initialized.

Notably, this isn't a brand-new crash class — the pre-fix code had the same latent crash via a different path (\`safe_confirm()\` → \`click.prompt()\` → \`RuntimeError: input(): lost sys.stdin\` under the identical closed-fd condition). But the whole point of the TTY fix was "non-interactive dango init should never crash or misbehave," so this gap in that same fix was worth closing rather than leaving as a known pre-existing issue.

Reproduced directly before and after the fix:
\`\`\`
$ python3 -c "import sys; print(sys.stdin.isatty())" 0<&-
AttributeError: 'NoneType' object has no attribute 'isatty'
\`\`\`

**Fix:** treat a \`None\` stdin or any exception from \`.isatty()\` the same as "not a TTY" — skip the prompt, persist nothing.

70 tests now in the two telemetry test files (up from 68), including two new regression tests for this exact scenario. Full suite: 4267 passed, 0 failed.

aaronteoh:
## Independent review found 2 issues that undermined this feature's entire purpose

A review was run against the fixed code and found two serious issues, both empirically confirmed by direct reproduction before any code changed (not just trusted on the reviewer's word):

### 1. Piped/scripted consent answers were silently dropped

The `sys.stdin.isatty()` gate added in the previous round (to fix the earlier TTY lock-in bug) was too broad — it returned before ever calling `safe_confirm()` for **any** non-TTY stdin, including a pipe with a genuine answer sitting in it. `echo yes | dango init` (or any Docker entrypoint / setup script piping an answer) got silently ignored: no prompt, nothing persisted, indistinguishable from success.

Reproduced directly:
```
$ echo -e "yes\n" | python3 -c "...._prompt_telemetry_consent(config)..."
# no prompt text, no ~/.dango/telemetry.json written at all
```
This is also why every other `safe_confirm()` call site in the codebase (checked all ~25) relies on click's own EOF handling instead of pre-checking `isatty()` — this was the only one doing something different, and it broke a real, common automation pattern.

**Fix:** replaced the `isatty()` pre-check with a dedicated `_ask_telemetry_consent()` helper that calls `click.prompt()` directly and returns `None` (instead of substituting a default) only when no real answer could be obtained at all — genuine EOF, or a closed-stdin exception. A real piped answer, TTY or not, is now read and honored normally. Verified: piping `"yes\n"` now correctly writes `{"enabled": true, "uuid": "..."}`.

### 2. The install ping essentially never landed on a real `dango init` exit

`ping()`'s daemon thread was never joined — a short-lived CLI process exits so fast that the thread gets killed before it can even build the payload.

Reproduced directly: 3/3 clean runs of `ping()` with **no artificial delay** wrote **no UUID to disk at all** (proving the thread's first line never executed), despite my earlier manual verification passing — because that test added a `time.sleep()` afterward to keep the process alive, which isn't representative of a real `dango init` exit. This meant the entire point of the session — making install counts measurable — likely wasn't actually working for real users.

**Fix:** `ping()` now joins the thread with a `_TIMEOUT_SECONDS` timeout instead of leaving it fully un-joined. Verified: the real HTTP round trip now reliably completes (`200`) within ~200ms under normal network conditions, and the wait is still capped at the timeout if the network hangs.

### Also fixed: dead `is_ci` payload field

`ping()` already refuses to run at all when `is_ci()` is true, so every payload that reaches the network call would always carry the same fixed `is_ci: false` — a field that can never vary was implying information it doesn't carry. Removed; the Worker's schema already defaults this column to `0` when absent, so this is a pure client-side simplification with no server-side behavior change.

---

**Note on verification noise:** while re-testing fix #2 against the live endpoint, I hit a few transient failures that initially looked like a deeper timing bug, traced them to Cloudflare's own rate limiting (the Worker's spec caps at >10 req/hour/IP) from the sheer volume of manual live-endpoint testing across this PR's review cycles — confirmed by the exact same test succeeding cleanly moments later. Also: one verification command was mistakenly run without an isolated `$HOME` override and briefly wrote a real `~/.dango/telemetry.json` on the local dev machine (anonymous UUID + OS/Python version + `source_types: ["csv"]`, no PII) and likely sent one real ping — caught immediately, file deleted, no lasting state.

84 tests now across the two telemetry test files (up from 70). Full suite: 4271 passed, 0 failed.

aaronteoh:
## Re-review response

Thanks for confirming fixes 1-3 hold up. Responding to the 3 new findings:

**#1 (PR description vs. \`not failures\`)** — verified you're right that the PR body still said "guarded by \`not failures and not skip_wizard\`" while the code only checks \`not skip_wizard\`. But I'd push back on the suggested direction: the fix here should be correcting the **description**, not re-adding \`not failures\` to the **code**. That guard was already confirmed dead code in round 2 (grepped: \`failures\` is never appended to anywhere in \`initialize()\`) and removed deliberately for that reason — re-adding it now would just reintroduce known-dead scaffolding to make a stale sentence true again, which is backwards. I rewrote the PR description instead to match the current, correct code (and while at it, caught that the description was stale in several other places too — test counts, the piped-stdin verification claim describing the now-replaced \`safe_confirm()\`-based mechanism, the line-count history). Full rewrite is up on the PR now.

**#2 (unlocked telemetry.json race)** — agreed, unchanged, still the same low-severity/accepted-risk assessment as originally documented. Left as-is, now explicitly listed as a known follow-up in the PR description.

**#3 (\`_ask_telemetry_consent()\` duplicates \`safe_confirm()\`'s loop)** — agreed this is real duplication, agreed it's not worth touching now. Unifying them would mean changing \`safe_confirm()\`'s shared contract (it currently returns a plain \`bool\`, always substituting a default rather than signaling "no answer obtained") across ~25 call sites for a benefit that's local to one caller. Listed as a follow-up in the PR description rather than done here.

No code changes from this round — just the PR description correction. Full suite still green (4271 passed).

aaronteoh:
## Final review — approved for merge

Summarizing where this PR landed after three independent review rounds, each of which caught something real:

**Round 1** (initial implementation): corrected a hook point that didn't exist in the codebase, fixed a config-shape error, and caught a live bug where Cloudflare's bot-fight mode silently 403'd every real ping (missing `User-Agent` header).

**Round 2**: fixed a permanent-opt-out lock-in bug, non-standard `DO_NOT_TRACK`/`DANGO_TELEMETRY` value matching, a CI-detection false-negative, a blocking network call, disabled sources leaking into the payload, and dead code in a guard condition — then a targeted follow-up caught that the fix for the lock-in bug introduced its own unguarded crash on closed stdin.

**Round 3**: found that the closed-stdin fix from round 2 was over-broad enough to silently drop legitimate piped answers (`echo yes | dango init`), and — more seriously — that the un-joined daemon thread from round 2's blocking fix meant the ping essentially never completed on a real, undelayed CLI exit. Both reproduced directly (not just claimed) before and after fixing. Also removed a dead `is_ci` payload field.

A fourth pass confirmed all of the above hold up under fresh reproduction, flagged a stale PR description (fixed — description now matches current code) and two accepted low-severity follow-ups (unlocked single-process file write, minor duplication between the new consent-prompt helper and the existing `safe_confirm()` utility), neither of which are correctness bugs.

**What I did before this review** to close the remaining gap between "the extracted method works" and "the real CLI works": ran `dango --help` / `dango init --help` through the actual worktree build to rule out import/wiring regressions, and ran a real `dango init --skip-wizard` subprocess end-to-end (not a mocked or direct method call) — clean exit, full project initialized, no telemetry prompt shown (correctly gated). The interactive-wizard path itself uses `inquirer`, which requires a real TTY and isn't newly introduced by this PR, so it wasn't re-tested here; the telemetry prompt logic downstream of it has been exercised directly and repeatedly with real (non-mocked) `click.prompt()` calls, real piped stdin, and real network round-trips against the live endpoint.

**Current state:** 84 tests across the two telemetry test files, 4271 passed / 0 failed in the full suite, all CI checks green (lint, mypy, tests on 3.10/3.12/3.13, security audit). Approved.

PR #442, opened 2026-09-01
1.0.8-D: Unified telemetry control

aaronteoh:
## Summary
- Add `dango telemetry` command group (`status` / `on` / `off`)
- Unified control: `--all` (or `--provider X`) controls Dango + dbt + dlt + Metabase in one command, writing through to each provider's own real config mechanism
- dbt: injects `DBT_SEND_ANONYMOUS_USAGE_STATS=false` + `DO_NOT_TRACK=1` into the subprocess env at **every** real dbt-invoking `subprocess.run` call site in the codebase (10 total — see "Independent review" below for how this list grew from 3 to 10)
- dlt: writes `dlthub_telemetry = false` (native TOML boolean) to `.dlt/config.toml`
- Metabase: calls `PUT /api/setting/anon-tracking-enabled` via `set_metabase_telemetry()` in `visualization/metabase.py`, reading the real `metabase_url` from `.dango/metabase.yml` and caching last-known state locally so `dango telemetry status` reflects reality
- Dango: delegates to `dango.telemetry` from Session C
- Updates `docs/network-egress.yml`'s `dango`/`metabase` `control` fields (previously placeholder text written before this session existed) to describe the real commands, and fixes a wrong field name (`allow_tracking` is the one-time `/api/setup` wizard field, not the runtime setting key)

**Wording:** every user-facing string says "controls telemetry for the components Dango configures" — never "nothing leaves your machine" (per task spec).

## Scope note: no web UI in this PR
The task file's title/Background mention a web settings page, but its Required Changes / Tests / Acceptance Criteria never actually specify any web route, template, or endpoint. Confirmed with the coordinating chat that this is a gap in how the task was originally scoped, not something to paper over — this PR is CLI-only. A follow-up task will cover the web settings page.

## Independent review — findings and how they were resolved
An independent multi-angle review (line-by-line diff scan, removed-behavior audit, cross-file call-site tracing, reuse/simplification/altitude passes) was run against this PR before merge, with every finding verified against the live worktree rather than trusted on report alone. Results, categorized:

**Fixed in this PR (blocking/high-severity):**
1. **dbt telemetry write-through only covered 3 of ~10 real dbt subprocess call sites.** The original PR wired `_dbt_telemetry_env()` into `run_dbt_models`/`run_dbt_snapshots`/`generate_dbt_docs` only. Two independent finder angles each found this from different directions; I personally verified all 7 additional sites by reading each one. Now wired into all of them: `cli/commands/transform.py` (`dango run`, `dango docs` — the primary user-facing dbt commands), `platform/local/watcher_runner.py` (file-watcher auto-triggered dbt run — previously silent since it's unattended), `web/routes/dbt.py` (web UI model-run endpoint), `cli/commands/dev.py` (`dango dev`), `cli/init.py` (`dango init`'s doc generation), `cli/model_wizard.py` (`dango model add/remove`), `cli/validate.py` (`dango validate`). This was the one that actually broke the "unified control" claim for most real dbt invocations — now fixed.
2. **`_get_metabase_telemetry_state()` was hardcoded `True`**, so `dango telemetry status` never reflected a real successful toggle. Fixed: `set_metabase_telemetry()` now writes a local last-known-state cache (`.dango/metabase_telemetry_state`) after every successful live API call; status reads that cache instead of guessing, with no live API call needed just to print status.
3. **`set_metabase_telemetry()` hardcoded `metabase_url="http://localhost:3000"`**, ignoring the real URL stored in `.dango/metabase.yml`. Fixed to read `credentials.get("metabase_url", "http://localhost:3000")`, matching the existing precedent in `cli/commands/metabase_cmd.py:343`.
4. **Uncaught exceptions in `set_metabase_telemetry()`** — only `click.ClickException`/`requests.exceptions.RequestException` were caught, so a malformed `metabase.yml` (`yaml.YAMLError`) or a non-JSON login response produced a raw traceback. Added a broad `except Exception` fallback converting anything else in the credentials/login/API flow to the same clean `click.ClickException` contract.
10. **Duplicate `creds_file.exists()` check** between `telemetry.py`'s wrapper and `metabase.py`'s `set_metabase_telemetry()` (byte-identical error message in both places — dead code in the first). Removed from `telemetry.py`'s wrapper; kept in `set_metabase_telemetry()` itself, since that's the function that actually reads the credentials file.

**Accepted trade-off — documented, not changed:**
5. `dango telemetry on --provider dbt` doesn't clear ambient `DO_NOT_TRACK`/`DBT_SEND_ANONYMOUS_USAGE_STATS` if the parent shell already exports them, so status can say "on" while dbt subprocesses still inherit an ambient opt-out. This is treated as correct behavior, not a bug: respecting an explicit ambient `DO_NOT_TRACK` (a user's own privacy signal, set outside Dango entirely) takes precedence over a Dango-level "on" the same way it takes precedence everywhere else in this codebase (`dango/telemetry.py`'s own `is_telemetry_enabled()` gives `DO_NOT_TRACK` the same override priority). Silently overriding a user's explicit ambient opt-out to force telemetry back on would be the wrong default for a privacy-focused command.
7. `.dlt/config.toml` read/write is hand-rolled with `tomlkit` in `telemetry.py` instead of reusing `CredentialManager.load_config()`/`save_config()` in `dango/config/credentials.py`, which already owns this file. Not changed here: the task spec this PR implements explicitly directed the tomlkit approach (matching dlt's own value-type behavior, verified against the installed dlt package), so this wasn't something introduced by choice — it's a legitimate follow-up (`CredentialManager.save_config()` uses the `toml` library, which does still serialize native booleans correctly, but loses `tomlkit`'s comment-preservation) rather than a defect in this PR.
8. Manual `project_root is None` checks in `telemetry.py` duplicate the codebase-wide `require_project_context()` helper. Not changed here: the `--all` flow deliberately treats a missing project as a soft skip-and-continue for dlt/metabase rather than a hard abort (see the acceptance criterion this implements), and `require_project_context()` always raises immediately — it can't be swapped in at the point this check lives without restructuring the soft-skip behavior it's specifically there to support.

**Deferred as follow-up tech debt (tracked in `ROADMAP.md` by the coordinating chat, not touched in this PR):**
6. Metabase auth flow (load creds → session → login → header) now duplicated three times across `setup_metabase()`, `sync_metabase_schema()`, and `set_metabase_telemetry()`, each with different error-handling conventions.
9. Bespoke `~/.dango/dbt_telemetry` plain-text sentinel file duplicates the existing `~/.dango/config.yml` machine-level config mechanism `dango/telemetry.py` already reads/writes — no other precedent for a bare sentinel file under `~/.dango/`.

## Corrections made vs. the task file's pseudocode
Grepped every path/class/function name against the live `v1.0.8` branch before implementing; found and fixed:
1. `dango.cli.utils.get_project_root_or_abort` doesn't exist (and was unused in the pseudocode anyway) — dropped.
2. `dango.config.helpers.load_config` import in the Metabase helper was unused — dropped.
3. **dbt env var name was wrong**: pseudocode used `DBT_ENGINE_SEND_ANONYMOUS_USAGE_STATS`; the real dbt-core var (verified against installed dbt-core 1.11.14 source and Session B's own `test_dbt_send_anonymous_usage_stats_envvar_name`) is `DBT_SEND_ANONYMOUS_USAGE_STATS`.
4. `MetabaseClient` class doesn't exist. The real class is `MetabaseProvisioner`, with no `from_project`/`set_setting`. Implemented `set_metabase_telemetry()` as a module-level function instead, matching the existing `sync_metabase_schema()` pattern.
5. dlt's `dlthub_telemetry` field is `bool`-typed, and dlt's own CLI writes a native TOML boolean — fixed from the pseudocode's string form.
6. **Acceptance-criteria bug**: the pseudocode's `--all` loop would abort entirely on the first project-scoped provider instead of warning and continuing. Fixed with per-provider error handling.
7. `docs/network-egress.yml`'s Metabase `control` field said `allow_tracking=false` — that's the one-time `/api/setup` wizard field name, not the runtime setting key (`anon-tracking-enabled`, confirmed against Metabase's public API docs). Fixed the doc wording.
8. Added try/except around the Metabase HTTP calls so a stopped Metabase raises a clean `ClickException`, not a raw `requests` exception (later broadened further — see review finding #4 above).

No lint/type/test check was resolved by adding to an exemption or skip mechanism — all corrections above are real fixes, not workarounds.

## Open question — left undisturbed, not resolved
Session B's egress test found that dlt's telemetry runtime never actually initializes for Dango's CSV-source sync path (`RunContext.runtime_config` is never touched), so there was nothing live to flip in that specific path. This PR's `dlthub_telemetry=false` write-through is correct and safe regardless — it matches dlt's own documented opt-out mechanism and takes effect if/when dlt's telemetry path ever does initialize. This PR does **not** claim the write-through has been observed to change real dlt network behavior — that remains an open question, documented as such in `docs/network-egress.yml`'s `notes` section.

## Verification
- `pytest -x`: 4300 passed, 30 skipped, 0 failed (full suite; +12 tests over the previous round)
- `pytest tests/unit/test_telemetry_unified.py -v`: 24 passed (12 original + 12 new, covering all 5 review fixes plus a structural regression guard asserting every known dbt subprocess call site passes `env=_dbt_telemetry_env()`)
- `mypy dango/`: Success, no issues found in 233 source files
- `ruff check dango/` / `ruff format --check dango/`: clean
- `python3 scripts/validate_headers.py` / `check_docstrings.py` on all changed files: clean
- Session B's egress tests, both classes: `pytest tests/unit/test_egress_allowlist.py -m "not egress"` (4 passed) and `-m egress` (1 passed) — both still pass
- **Live CLI runs**, isolated via `HOME=` override (never touched the real machine's `~/.dango/`):
  ```
  $ dango telemetry off --all   # outside a project
  ✓ dango: telemetry disabled
  ✓ dbt: telemetry disabled
  ! dlt: skipped — not inside a Dango project
  ! metabase: skipped — not inside a Dango project

  $ dango telemetry off --all   # inside a minimal project, Metabase not configured
  ✓ dango: telemetry disabled
  ✓ dbt: telemetry disabled
  ✓ dlt: telemetry disabled
  ! metabase: skipped — Metabase not configured. Run dango start first.

  $ cat .dlt/config.toml
  [runtime]
  dlthub_telemetry = false

  $ dango telemetry status   # no metabase.yml, no cache file — clean default, no crash
  metabase │ on │ Anonymous usage statistics │ dango telemetry on/off --provider metabase
  ```
  Metabase's live API call path is covered by mocked unit tests only — no running Metabase in this worktree session per the "don't test on shared projects" rule.

## Regression check
- dbt subprocess env verified: `_dbt_telemetry_env()` returns `{**os.environ}` copy, never mutates `os.environ` in place (explicit test: `test_dbt_env_is_a_copy_not_the_real_environ`)
- Metabase call confirmed graceful on connection failure, malformed YAML, and non-JSON login response (all convert to `click.ClickException`, never a raw traceback)
- `--all` outside a project succeeds for dango+dbt, warns (not crashes) for dlt+metabase — verified both by unit test and live CLI run above
- `dango telemetry status` now reflects a real Metabase toggle end-to-end (previously always showed "on") — verified by `test_status_reflects_real_toggle_end_to_end`

## Second narrower review pass — findings and fixes
A follow-up review scoped specifically to the changes above (the Metabase state-cache write/read, the 8 newly-wired dbt call sites, and the new `except Exception` handler) — not a full re-audit — found and fixed two more real issues:

1. **`_set_dbt_telemetry()`/`_set_dlt_telemetry()` raised unwrapped `OSError`/`tomlkit.exceptions.TOMLKitError`** instead of `click.ClickException`, breaking the `--all` graceful-degradation contract every other provider follows (a read-only `~/.dango`, or a hand-corrupted `.dlt/config.toml`, would crash the whole `--all` command instead of skipping that provider). Both now wrap their filesystem/tomlkit operations and convert to `click.ClickException`.
2. **A Metabase login 401/403 (stale admin password) was mislabeled "Could not reach Metabase — is it running?"** since `requests.HTTPError` is a `RequestException` subclass. Metabase is running fine in this case — the credentials are wrong. Now checked explicitly on the login response status code, with a distinct "check admin credentials in .dango/metabase.yml" message.

Also proactively found and fixed, ahead of that review pass, by re-reading the previous round's own fix: the Metabase state-cache write (`.dango/metabase_telemetry_state`) landed inside the same `try` as the real API call, so a failure writing that *secondary* cache file (disk full, permissions) would report a *successful* telemetry toggle as a command failure. Fixed: the cache write is now wrapped in its own try/except that logs a warning and returns normally on failure — real API success is never reported as failure due to the secondary write.

`tests/unit/test_telemetry_unified.py` grew to 568 lines during this round, tripping the 500-line file-size-check pre-commit hook. Split along its existing class boundaries into `test_telemetry_unified.py` (dbt/dlt + top-level CLI tests) and a new `test_telemetry_unified_metabase.py` (all Metabase-specific tests) — not exempted, since this is new code from this PR, not legacy debt.

## Final verification (after all review rounds)
- `pytest -x`: 4305 passed, 30 skipped, 0 failed
- `pytest tests/unit/test_telemetry_unified.py tests/unit/test_telemetry_unified_metabase.py -v`: 29 passed
- `mypy dango/`: Success, no issues found in 233 source files
- `ruff check` / `ruff format --check`: clean
- header/docstring/file-size checks: clean
- Session B's egress tests (both classes): still pass

### DIFF
--- a/dango/cli/CLAUDE.md
+++ b/dango/cli/CLAUDE.md
@@ -47,6 +47,7 @@ Click-based command-line interface for all Dango operations — project init, so
 | `commands/analyze.py` (~97 lines) | `monitor` group + `analyze` alias | `monitor` (group), `monitor_run()`, `analyze()` |
 | `commands/dev.py` (~429 lines) | `dev` group (default run + clean) — branch-based dbt development | `dev` (group), `dev_clean()` |
 | `commands/web.py` (69 lines) | `web` dev server command | `web()` |
+| `commands/telemetry.py` (~250 lines) | `telemetry` group (status, on, off) — unified write-through control for Dango, dbt, dlt, and Metabase telemetry | `telemetry`, `telemetry_status()`, `telemetry_on()`, `telemetry_off()` |
 | **Wizards** | | |
 | `init.py` (1585 lines) | Project initialization wizard, incl. first-run telemetry consent prompt | `ProjectInitializer` |
 | `wizard.py` (307 lines) | Interactive setup wizards | `ProjectWizard` |
@@ -136,8 +137,12 @@ dango (top-level group)
 │   ├── accept, drift-report, pii-report, pii-set, pii-list
 ├── notebook (group)            ← commands/notebook.py
 │   ├── new, open              (default invocation lists notebooks)
-└── metabase (group)            ← commands/metabase_cmd.py
-    ├── save, load, refresh
+├── metabase (group)            ← commands/metabase_cmd.py
+│   ├── save, load, refresh
+└── telemetry (group)           ← commands/telemetry.py
+    ├── status
+    ├── on   [--all | --provider dango|dbt|dlt|metabase]
+    └── off  [--all | --provider dango|dbt|dlt|metabase]
 ```
 
 ## Seeds vs CSV Data Source
--- a/dango/cli/commands/dev.py
+++ b/dango/cli/commands/dev.py
@@ -121,7 +121,7 @@ def _run_dev_dbt(project_root: Path, dev_dir: Path, select: str | None) -> int:
     """
     import subprocess
 
-    from dango.transformation import _get_dbt_executable
+    from dango.transformation import _dbt_telemetry_env, _get_dbt_executable
 
     dbt_cmd = _get_dbt_executable()
     dbt_dir = project_root / "dbt"
@@ -133,7 +133,7 @@ def _run_dev_dbt(project_root: Path, dev_dir: Path, select: str | None) -> int:
     console.print(f"[dim]Running: {' '.join(cmd)}[/dim]\n")
 
     try:
-        result = subprocess.run(cmd, cwd=str(dbt_dir), timeout=300)
+        result = subprocess.run(cmd, cwd=str(dbt_dir), timeout=300, env=_dbt_telemetry_env())
         return result.returncode
     except subprocess.TimeoutExpired:
         console.print(
--- a/dango/cli/commands/telemetry.py
+++ b/dango/cli/commands/telemetry.py
@@ -173,10 +173,21 @@ def _set_dbt_telemetry(enabled: bool) -> None:
     Machine-level (~/.dango/), matching Dango's own telemetry identity scope
     (see dango/telemetry.py) — one opt-out covers every project on the
     machine.
+
+    Raises:
+        click.ClickException: If the sentinel file can't be written (e.g.
+            ~/.dango is read-only or otherwise inaccessible) — converted
+            from a raw OSError so `--all` can skip this provider and
+            continue with the rest instead of crashing the whole command,
+            matching the error-handling contract set_metabase_telemetry()
+            uses for every one of its own failure modes.
     """
     sentinel = Path.home() / ".dango" / "dbt_telemetry"
-    sentinel.parent.mkdir(parents=True, exist_ok=True)
-    sentinel.write_text("true" if enabled else "false")
+    try:
+        sentinel.parent.mkdir(parents=True, exist_ok=True)
+        sentinel.write_text("true" if enabled else "false")
+    except OSError as e:
+        raise click.ClickException(f"Could not write dbt telemetry sentinel: {e}") from e
 
 
 def _set_dlt_telemetry(enabled: bool, project_root: Path | None) -> None:
@@ -185,18 +196,33 @@ def _set_dlt_telemetry(enabled: bool, project_root: Path | None) -> None:
     Writes a native TOML boolean (not a string) to match the value type
     dlt's own `dango telemetry`-equivalent CLI (`dlt telemetry switch`)
     writes — RuntimeConfiguration.dlthub_telemetry is a bool-typed field.
+
+    Raises:
+        click.ClickException: If project_root is None, if the file can't
+            be read/written (OSError — e.g. permissions), or if the
+            existing .dlt/config.toml is malformed
+            (tomlkit.exceptions.TOMLKitError, e.g. from manual editing) —
+            the latter two converted from raw exceptions so `--all` can
+            skip this provider and continue, same contract as dbt above
+            and Metabase's set_metabase_telemetry().
     """
     import tomlkit
+    from tomlkit.exceptions import TOMLKitError
 
     if project_root is None:
         raise click.ClickException("Must be run inside a Dango project for dlt control")
     config_path = project_root / ".dlt" / "config.toml"
-    config_path.parent.mkdir(parents=True, exist_ok=True)
-    doc = tomlkit.parse(config_path.read_text()) if config_path.exists() else tomlkit.document()
-    if "runtime" not in doc:
-        doc.add("runtime", tomlkit.table())
-    doc["runtime"]["dlthub_telemetry"] = enabled
-    config_path.write_text(tomlkit.dumps(doc))
+    try:
+        config_path.parent.mkdir(parents=True, exist_ok=True)
+        doc = tomlkit.parse(config_path.read_text()) if config_path.exists() else tomlkit.document()
+        if "runtime" not in doc:
+            doc.add("runtime", tomlkit.table())
+        doc["runtime"]["dlthub_telemetry"] = enabled
+        config_path.write_text(tomlkit.dumps(doc))
+    except OSError as e:
+        raise click.ClickException(f"Could not write .dlt/config.toml: {e}") from e
+    except TOMLKitError as e:
+        raise click.ClickException(f"Could not parse .dlt/config.toml: {e}") from e
 
 
 def _set_metabase_telemetry(enabled: bool, project_root: Path | None) -> None:
--- a/dango/cli/commands/telemetry.py
+++ b/dango/cli/commands/telemetry.py
@@ -73,7 +73,7 @@ def telemetry_status(ctx: click.Context) -> None:
     )
 
     # Metabase
-    mb_on = _get_metabase_telemetry_state()
+    mb_on = _get_metabase_telemetry_state(project_root)
     table.add_row(
         "metabase",
         "[green]on[/green]" if mb_on else "[dim]off[/dim]",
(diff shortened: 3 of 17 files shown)


aaronteoh:
## Independent review — final pass

Did a fresh, holistic read of the complete diff as it stands now (not just the incremental pieces from each fix round) before posting this: `dango/cli/commands/telemetry.py`, `dango/visualization/metabase.py`, `dango/transformation/__init__.py`, and `docs/network-egress.yml` in full. Everything is internally consistent — docstrings match behavior, the egress doc matches the code, and the wording constraint ("controls telemetry for the components Dango configures," never "nothing leaves your machine") holds everywhere I can find.

**Process this PR went through:**
- Round 1: implemented against the task spec, correcting 8 verified discrepancies between the spec's pseudocode and the live codebase (wrong dbt env var name, nonexistent `MetabaseClient` class, wrong dlt value type, an acceptance-criteria bug in the `--all` loop, a wrong Metabase setting-key claim in the docs, etc.) — all caught by grepping the actual codebase before writing code, not assumed from the spec.
- Round 2 (first independent review): found the dbt telemetry write-through only covered 3 of ~10 real dbt subprocess call sites — the one genuinely blocking finding, since it undercut the PR's core "unified control" claim for most everyday dbt invocations (`dango run`, the file-watcher's auto-run, the web UI's model-run button, etc.). Fixed, plus three smaller but real Metabase issues (status hardcoded rather than reflecting real state, URL hardcoded rather than reading stored config, uncaught exception classes). All fixes carry regression tests, including a structural guard asserting every known dbt call site passes `env=_dbt_telemetry_env()`.
- Proactive: re-reading round 2's own fix surfaced a failure-ordering bug (a failed *secondary* state-cache write could report a *successful* API call as a command failure) — fixed ahead of the next review round rather than left for it to rediscover.
- Round 3 (second independent review, scoped to just the new code): found two more real issues — `_set_dbt_telemetry`/`_set_dlt_telemetry` could raise unwrapped `OSError`/`TOMLKitError` instead of `click.ClickException`, breaking `--all`'s graceful-skip contract; and a Metabase 401/403 login failure was mislabeled "is it running?" instead of a credentials problem. Both fixed and tested.

**Deliberately out of scope, documented as such:**
- No web UI in this PR — the task file's Required Changes/Tests/Acceptance Criteria never specified one despite the title mentioning it; confirmed with the coordinating chat this is a gap in the task's own scoping, not something to silently fill. Follow-up task planned separately.
- dlt's write-through has not been observed to change real dlt network behavior (Session B found dlt's telemetry runtime never initializes for the CSV-source path) — left as a documented open question in `docs/network-egress.yml`, not resolved, per the task's own optional framing.
- Three items accepted as deliberate trade-offs with reasoning stated in the PR body: ambient `DO_NOT_TRACK` taking precedence over an explicit `dango telemetry on --provider dbt`, the tomlkit-vs-`CredentialManager` duplication (task-spec-directed, not introduced by choice), and the `require_project_context()` duplication (the `--all` soft-skip design needs the check at that level).
- Two items deferred as tracked tech debt (Metabase auth flow duplicated 3x, bespoke dbt sentinel file vs. `config.yml`) — coordinating chat is adding these to `ROADMAP.md` after merge.
- One out-of-PR finding surfaced by a review methodology slip (diffed against `main` instead of `v1.0.8`, pulling in already-merged Session C code): a real bug in `dango/telemetry.py`'s `is False` identity check, but not this PR's file — flagged to the coordinating chat to decide on separately, not touched here.

**Current state:** `pytest -x` 4305 passed / 30 skipped / 0 failed, `mypy dango/` clean (233 files), `ruff check`/`ruff format --check` clean, header/docstring/file-size checks clean, Session B's egress tests (both classes) still pass, CI green across all required and informational checks. No lint/type/test check was ever resolved by adding to an exemption or skip mechanism — every fix in this PR is a real fix, including a mid-review test-file split (not an exemption) when a test file grew past the 500-line limit.

I'm confident this is in good shape to merge.

PR #443, opened 2026-09-02
1.0.8-E: MCP server — scaffold + read tools + setup

aaronteoh:
## Summary
- Add `dango mcp` command group: `run` / `setup` / `status`
- MCP server (fastmcp, local stdio only) exposes 8 read tools: `list_sources`,
  `get_table_schema`, `get_catalog`, `get_lineage`, `list_models`, `get_model_sql`,
  `query`, `get_sync_history`
- `dango mcp setup` auto-configures Claude Code, Cursor, Windsurf
- `dango mcp status` verifies configuration
- All Dango imports lazy inside tool functions (no stdout pollution — the server
  communicates over stdio, so any stray stdout write corrupts the JSON-RPC protocol)

## Deviations from the session spec, found via grep before implementing
Per the task's own instruction to verify pseudocode against the live code (this
project has a documented history of task files with wrong method/function names),
I found and fixed 4 real bugs beyond the phantom import already removed pre-dispatch:

1. **`get_sync_history`** called a nonexistent
   `dango.utils.sync_history.get_recent_history`. Rewritten against the real
   `load_sync_history(project_root, source_name, limit)`; aggregates across all
   configured sources when `source_name` is omitted (no native "all sources" mode
   exists in the underlying function).
2. **`list_sources`** called `get_source_status_data(source.name, project_root=...)`.
   That function actually takes a source **dict** (not a name string), has **no
   `project_root` kwarg**, and returns a pydantic `SourceStatus` object (not a
   dict). Worse: it — and everything else in `web/helpers.py` — resolves the
   project root via `dango.web.app.app.state.project_root`, which is only
   populated by FastAPI's `lifespan()` when the real web server boots. A
   standalone MCP process never boots that app, so this would raise `RuntimeError`
   on every call, silently degrading to `status: "unknown"` for every source
   (swallowed by the original pseudocode's bare `except Exception`). Fixed by
   using `load_sources_config()` (dict-based, matches the real
   `web/routes/sources.py` call site) and setting `app.state.project_root`
   before calling — confirmed safe re: stdout (only `lifespan()` prints, never
   `create_app()`/import time).
3. **`query()`'s SELECT-only check** was a bare `startswith("SELECT")`, which
   rejects legitimate `WITH ... SELECT` CTEs and doesn't block multi-statement
   injection (`SELECT 1; DROP TABLE x`). Replaced with the same sqlglot-AST
   validation (falling back to a keyword + multi-statement check when sqlglot
   is unavailable) that `web/routes/query.py`'s `_validate_sql` already uses.
4. **Read-only DuckDB connections aren't unconditionally safe** against a
   concurrent writer, contra the spec's regression-risk note. Added the same
   retry-on-`IOException` (3x, exponential backoff) that
   `web/routes/query.py`'s `_execute_query` already uses, to `query()`,
   `get_table_schema()`, and `get_catalog()`.

Also: `mcp.run(show_banner=False)` — fastmcp's default banner triggers a
blocking network call to `https://pypi.org/pypi/fastmcp/json` on every server
start (to check for a newer fastmcp version). That's an unwanted, un-opt-in-able
outbound call on the local-stdio-only surface this session's egress dependency
gate (Session B) exists to protect, firing on every LLM-client-spawned session.
Disabled it; confirmed via the full test suite (including
`tests/unit/test_egress_allowlist.py`) that no new undisclosed host is contacted.

## live PR diff on GitHub)
10 findings came back. Each was evaluated individually rather than applied
wholesale:

**Fixed (6):**
1. `mcp_setup.py`: `sys.executable.replace("/bin/python", "/bin/dango")` broke
   on this repo's own documented `python3.11 -m venv venv` setup (leaves a
   trailing "3.11" on the guessed path, e.g. `.../bin/dango3.11`, which doesn't
   exist) and was a complete no-op on Windows (no `/bin/python` substring to
   replace, so it would write `python.exe` itself as the command — broken with
   no fallback triggered). Replaced with `Path(sys.executable).parent /
   "dango"` — console scripts always live next to the interpreter — falling
   back to bare `"dango"` on PATH only when no sibling script exists.
2. `get_table_schema()`: omitting `schema` on a table name that exists in more
   than one schema (e.g. a generic name reused across two raw sources)
   silently merged every matching table's columns into one list under a
   single misleading schema name. Now filters to the first (alphabetical)
   schema's columns only, and surfaces the ambiguity via an `other_schemas`
   list + `note` field rather than hiding it.
3. `query()` had no execution timeout, unlike `web/routes/query.py`'s
   `asyncio.wait_for(timeout=query_timeout)`. Turned out cheap to fix once
   the DB-execution logic was already being extracted for the length-cap fix
   below: `query()` is now `async def`, wrapping
   `asyncio.to_thread(_execute_query_sync, ...)` in `asyncio.wait_for(...,
   timeout=30)` (matching the web endpoint's default), returning a clean
   error instead of hanging the server on an expensive query.
4. `_write_mcp_config()` did a direct `write_text()` onto the user's real LLM
   client config file (`~/.claude/settings.json` etc.) with no atomicity — an
   interrupted write (crash, kill, laptop sleep) could corrupt the file
   beyond just the MCP section, including unrelated settings. Now writes to a
   temp file in the same directory and `os.replace()`s it in.
5. `get_lineage()` manually reconstructed a model's `unique_id` as
   `f"model.{package_name}.{model_name}"`, which breaks for dbt's versioned
   models (dbt-core 1.5+, and this repo pins 1.10/1.11) — the real
   `unique_id` has a `.vN` suffix the reconstruction omits, silently making
   `referenced_by` empty even when a real dependent exists. Now uses the
   manifest's own dict key directly.
6. `list_sources()` was a sync tool calling `asyncio.run()` internally, which
   only avoided "cannot be called from a running event loop" because
   fastmcp's default dispatches sync tools to a worker thread
   (`run_in_thread=True`) — an undocumented dependency that would silently
   break if that default ever changed. Converted to a native `async def`
   tool that awaits directly on fastmcp's own event loop.

   Also added: a length cap on `query()`'s `sql` parameter (102,400 chars,
   matching `web/routes/query.py`'s `QueryRequest.sql` field limit) — no prior
   upfront rejection of an oversized string before it reached sqlglot/DuckDB.

   Fixing #3 and #6 together made `mcp_server.py` async-heavier and pushed it
   back over the 500-line file-size check (518 lines). Did a second real
   split: extracted the plain-function helpers (`_get_project_root`,
   `_validate_select_only`, `_connect_readonly_with_retry`,
   `_execute_query_sync`, `_infer_layer`) into a new
   `dango/cli/commands/mcp_helpers.py`. This is a *different* kind of split
   from `mcp_setup.py` — no Click decorators involved, just a normal helper
   extraction imported at the top of `mcp_server.py` (that one eager import is
   safe: `mcp_helpers.py` itself does zero `dango.*` imports at its own module
   level — all real work stays lazy inside its function bodies). Final sizes:
   `mcp_server.py` 428 lines, `mcp_setup.py` 142, `mcp_helpers.py` 119.

**Accepted trade-off, not fixed (1):**
- No audit logging on `query()`, unlike `web/routes/query.py`'s
  `log_auth_event(AuditEvent.QUERY_EXECUTED, ...)` after every execution. The
  MCP server has no authenticated-user concept to log against — it's a single
  local operator's own coding agent, not a multi-tenant web app with a `User`
  to attribute the read to. `log_auth_event`'s shape (user_id, email, ip)
  doesn't map cleanly onto that. Left as-is; flagging here rather than
  silently dropping the finding.

**Rejected, not bugs (2):**
- Raw `str(e)` in every tool's error return "violates" STANDARDS.md's
  `HTTPException(detail=str(e))` prohibition. Disagreed: that rule targets
  HTTP APIs exposed to less-trusted callers. An MCP tool returning the real
  error text to the calling agent is the standard, useful pattern here — the
  agent needs the actual DuckDB/validation error to self-correct — and every
  one of the 8 tools does this uniformly, matching the spec's own pseudocode
  throughout. Different threat model, not a violation.
- `row_limit` hardcoded at 500 instead of honoring `config.api.query_max_rows`.
  This is explicitly spec-mandated (task's "Critical code patterns — do NOT
  violate": *"query() row_limit capped at 500 — do not allow unlimited
  queries"*, and it's a required acceptance-criteria test), not an oversight.

6 new regression tests added for the fixes above, including positive controls
for the two highest-severity bugs (`get_table_schema`'s column merge — two
same-named tables in different schemas — and `sys.executable`'s broken
venv-path guess — a real `pythonX.Y`-named fake interpreter with a sibling
console script) rather than just re-reading the fixed source.

## File split (not in the original single-file plan)
`mcp_server.py` hit the repo's 500-line pre-commit file-size check twice —
once at 560 lines from the initial bug fixes, again at 518 lines from the
review-round fixes above. Final structure:
- `dango/cli/commands/mcp_server.py` (428 lines) — FastMCP instance, all 8
  read tools, `mcp_group` + `run` subcommand.
- `dango/cli/commands/mcp_setup.py` (142 lines) — `setup`/`status`
  subcommands + `_write_mcp_config`, self-registering onto `mcp_group` via
  `@mcp_group.command()`, triggered by a bottom-of-file import in
  `mcp_server.py` — mirrors the existing cross-file registration pattern in
  `commands/remote.py` + `remote_auth.py`/`remote_mgmt.py`/etc.
- `dango/cli/commands/mcp_helpers.py` (119 lines) — plain-function helpers
  used by the read tools (no Click decorators, a plain import not a
  registration-triggering one).

No behavior or CLI-surface change from any of this; `main.py` still only
imports `mcp_group`.

## Verification
- `pytest -x` (foreground, before and after each review round, and again
  after rebasing onto `v1.0.8` post-Session-D-merge): last run **4339
  passed, 30 skipped, 0 failed**
- `ruff check dango/` / `ruff format --check dango/`: clean
- `mypy dango/cli/commands/mcp_server.py dango/cli/commands/mcp_setup.py
  dango/cli/commands/mcp_helpers.py dango/cli/main.py`: no issues
- `dango mcp --help` shows `run` / `setup` / `status`
- `dango mcp setup` / `dango mcp status` — ran twice with an isolated `HOME=`
  override (never touched this session's real `~/.claude/settings.json`);
  confirmed the fixed venv-path resolution writes the actual console script
  path (`.../venv/bin/dango`), not a broken guess
- Real subprocess stdio smoke test, run both before and after the
  review-round fixes: spawned `dango mcp run` as an actual child process and
  connected with `fastmcp.Client` over real stdio transport — `list_tools()`
  returns all 8 tools, and `list_sources()`/`query(...)` (now `async def`
  tools) execute correctly through the real JSON-RPC protocol, confirming no
  stdout corruption and no regression from the sync-to-async conversion

## already covered in the first round)
6 findings, each verified empirically (not just re-read) before deciding:

**Fixed (4):**
1. `query()`'s timeout via `asyncio.wait_for(asyncio.to_thread(...))` never
   actually cancelled the running query. Verified empirically (a standalone
   repro script) that the worker thread keeps executing to completion in the
   background after the tool call has already returned a timeout error to
   the caller, holding a slot in the shared default thread-pool executor the
   whole time — directly contradicting the previous round's docstring claim
   that queries are "killed" after the timeout. `query()` now creates its
   own connection, keeps an `asyncio.shield()`-protected handle to the
   execution task, and calls `conn.interrupt()` (DuckDB's own thread-safe
   query cancellation — confirmed present via `hasattr`/live call on duckdb
   1.5.4) when the timeout fires, then waits briefly for the worker to
   actually unwind before closing the connection, rather than closing
   underneath a still-running thread.
2. `_write_mcp_config`'s atomic write (added last round) silently downgrades
   the LLM client config file's permissions to 0600 on every `dango mcp
   setup` run. Verified empirically: `tempfile.mkstemp()` always creates its
   temp file at mode 0600 regardless of the target's prior mode — a
   pre-existing 0644 `settings.json` became 0600 after `os.replace()`. Now
   preserves the original file's mode (or defaults to 0644 for a new file).
3. `_QUERY_TIMEOUT_SECONDS` was hardcoded at 30, but the web query endpoint
   it claimed to match reads a per-project configurable
   `config.api.query_timeout_seconds`. Added `_get_query_timeout_seconds()`
   so the MCP tool honors the same setting a user may have customized, not
   just the same default value.
4. Duplicate `import asyncio` inside two function bodies instead of one
   module-level import — `asyncio` is stdlib and never touched the "lazy
   `dango.*` imports only" rule the duplication seemed to be guarding
   against. Consolidated to a top-level import.

**Resolved as a side effect of fix 1 (not a separate change):** the review
also flagged that `_execute_query_on_connection`'s internal call to
`_connect_readonly_with_retry` (both defined in `mcp_helpers.py`) was only
patchable via `mcp_helpers.X` in a test, not the `mcp_server.X` names every
other helper uses — verified empirically that
`monkeypatch.setattr(mcp_server, "_connect_readonly_with_retry", fake)` had
zero effect on `query()`'s actual DB connection. Restructuring `query()` to
own the connection lifecycle itself (fix 1) means
`_execute_query_on_connection` no longer calls `_connect_readonly_with_retry`
at all, so this inconsistency no longer exists on the `query()` path.
Documented the general pattern in `mcp_helpers.py`'s module docstring so a
future helper doesn't reintroduce it.

**Reported, not fixed (1):** `_write_mcp_config`'s mkstemp+os.replace atomic
write duplicates an existing, near-identical implementation in
`dango/governance/pii_overrides.py::_save_overrides` (same pattern, same
latent 0600-permission bug, unfixed there). Extracting a shared helper would
mean touching an unrelated module this PR doesn't otherwise change —
flagging as a follow-up candidate rather than doing that refactor here.

4 new regression tests, including positive controls for both fixed bugs: a
connection-wrapper spy proving `conn.interrupt()` is actually called on
timeout (duckdb's connection type doesn't allow monkeypatching individual
methods directly, since it's a C extension type), and a test proving
`settings.json`'s file mode survives a `dango mcp setup` run unchanged.

## Regression check
- `query()` confirmed SELECT-only (including CTE + multi-statement +
  oversized-SQL cases, unit tested), with a configurable execution timeout
  that now actually interrupts the in-flight query
- DuckDB connections confirmed `read_only=True` with retry-on-lock
- No stdout output before/during `mcp.run()` confirmed via real subprocess
  test, re-run after each round of fixes
- Session D (`feat/108-telemetry-unified`) merged into `v1.0.8` (PR #442)
  partway through this session. Fetched `v1.0.8`, confirmed the only overlap
  was `main.py` (each side adds one non-overlapping import + one
  `cli.add_command(...)` line, as the task spec predicted), and rebased
  cleanly — `git rebase origin/v1.0.8` completed with no conflicts. Re-ran
  the full verification suite (ruff/mypy/`pytest -x` foreground: 4339
  passed, 0 failed) and the real subprocess stdio smoke test after the
  rebase before force-pushing.

## Not merged
Per workflow: this PR is left for the coordinating chat to review and merge.

### DIFF
--- a/dango/cli/CLAUDE.md
+++ b/dango/cli/CLAUDE.md
@@ -27,6 +27,8 @@ Click-based command-line interface for all Dango operations — project init, so
 | `commands/model.py` (258 lines) | `model` group (`add`, `remove`) | `model` |
 | `commands/seed.py` (135 lines) | `seed` group (`add`, `list`) — dbt seed CSV management | `seed` |
 | `commands/dashboard.py` (153 lines) | `dashboard` group (`provision`) | `dashboard` |
+| `commands/mcp_server.py` (~470 lines) | `mcp` group + `run` command; FastMCP stdio server with 8 read-only tools (list_sources, get_table_schema, get_catalog, get_lineage, list_models, get_model_sql, query, get_sync_history) | `mcp_group`, `mcp`, `mcp_run()` |
+| `commands/mcp_setup.py` (~120 lines) | `mcp setup` / `mcp status` subcommands (registers onto `mcp_group` from mcp_server.py) — detects/configures Claude Code, Cursor, Windsurf | `mcp_setup()`, `mcp_status()` |
 | `commands/remote.py` (702 lines) | `remote` group → `push`, `rollback`, `firewall`, `domain` subgroups + management commands | `remote`, `remote_push()`, `remote_rollback()`, `firewall`, `domain` |
 | `commands/migrate.py` | `migrate` group (`status`, `run`) | `migrate` |
 | `commands/serve.py` (~205 lines) | `serve` production foreground server with `--workers` option. Metabase setup failure is non-fatal (prints warning, continues to uvicorn). | `serve()` |
@@ -99,6 +101,9 @@ dango (top-level group)
 │   ├── add, list
 ├── dashboard (group)           ← commands/dashboard.py
 │   ├── provision
+├── mcp (group)                 ← commands/mcp_server.py
+│   ├── run                     ← commands/mcp_server.py (FastMCP stdio server)
+│   └── setup, status           ← commands/mcp_setup.py
 ├── migrate (group)             ← commands/migrate.py
 │   ├── status, run
 ├── remote (group)              ← commands/remote.py
@@ -165,7 +170,8 @@ dango (top-level group)
 - **Lazy imports:** All command functions use `from dango.xxx import ...` inside the function body, not at module level. This prevents circular imports and speeds up CLI startup.
 - **Shared console:** All command modules import `console` from `dango.cli` for Rich terminal output.
 - **Project root via context:** `ctx.obj["project_root"]` is set by the top-level `cli` group in `main.py` (via `dango.config.helpers.find_project_root`).
-- **Cross-file command registration:** `remote_auth.py`, `remote_mgmt.py`, `remote_ops.py`, `remote_env.py`, `remote_backup.py`, `remote_repair.py`, and `remote_sync.py` import `remote` from `remote.py` and register commands via `@remote.command()` / `@remote.group()`. Registration is triggered by bottom-of-file imports in `remote.py`.
+- **Cross-file command registration:** `remote_auth.py`, `remote_mgmt.py`, `remote_ops.py`, `remote_env.py`, `remote_backup.py`, `remote_repair.py`, and `remote_sync.py` import `remote` from `remote.py` and register commands via `@remote.command()` / `@remote.group()`. Registration is triggered by bottom-of-file imports in `remote.py`. Same pattern for `mcp`: `mcp_setup.py` imports `mcp_group` from `mcp_server.py` and registers `setup`/`status` via `@mcp_group.command()`; triggered by a bottom-of-file import in `mcp_server.py`. Split out of a single file to stay under the 500-line file-size check.
+- **MCP tool functions are plain functions, not wrapped:** `@mcp.tool()` (fastmcp) returns the original function unchanged — call `mcp_server.list_sources()` etc. directly in tests, no `.fn` unwrapping needed. All `dango.*` imports inside MCP tool bodies must stay lazy (never at module top level) — the server communicates over stdio and any stray stdout write corrupts the JSON-RPC protocol.
 - **Two SSH users:** `root` for system ops (backup, rollback, domain, server setup) via `load_cloud_config_with_ssh()` in `cli/utils.py`. `dango` for project file ops (.env, .dlt/secrets.toml) via `_ssh_connect_or_fail()` in `remote.py`.
 
 ### Key conventions
--- a/dango/cli/CLAUDE.md
+++ b/dango/cli/CLAUDE.md
@@ -27,8 +27,9 @@ Click-based command-line interface for all Dango operations — project init, so
 | `commands/model.py` (258 lines) | `model` group (`add`, `remove`) | `model` |
 | `commands/seed.py` (135 lines) | `seed` group (`add`, `list`) — dbt seed CSV management | `seed` |
 | `commands/dashboard.py` (153 lines) | `dashboard` group (`provision`) | `dashboard` |
-| `commands/mcp_server.py` (~470 lines) | `mcp` group + `run` command; FastMCP stdio server with 8 read-only tools (list_sources, get_table_schema, get_catalog, get_lineage, list_models, get_model_sql, query, get_sync_history) | `mcp_group`, `mcp`, `mcp_run()` |
-| `commands/mcp_setup.py` (~120 lines) | `mcp setup` / `mcp status` subcommands (registers onto `mcp_group` from mcp_server.py) — detects/configures Claude Code, Cursor, Windsurf | `mcp_setup()`, `mcp_status()` |
+| `commands/mcp_server.py` (~420 lines) | `mcp` group + `run` command; FastMCP stdio server with 8 read-only tools (list_sources, get_table_schema, get_catalog, get_lineage, list_models, get_model_sql, query, get_sync_history) | `mcp_group`, `mcp`, `mcp_run()` |
+| `commands/mcp_setup.py` (~145 lines) | `mcp setup` / `mcp status` subcommands (registers onto `mcp_group` from mcp_server.py) — detects/configures Claude Code, Cursor, Windsurf | `mcp_setup()`, `mcp_status()` |
+| `commands/mcp_helpers.py` (~120 lines) | Plain-function helpers for the MCP read tools (project root, SELECT-only SQL validation, retry-on-lock DuckDB connect, blocking query execution) — split out of mcp_server.py, not a Click module | `_get_project_root()`, `_validate_select_only()`, `_connect_readonly_with_retry()`, `_execute_query_sync()` |
 | `commands/remote.py` (702 lines) | `remote` group → `push`, `rollback`, `firewall`, `domain` subgroups + management commands | `remote`, `remote_push()`, `remote_rollback()`, `firewall`, `domain` |
 | `commands/migrate.py` | `migrate` group (`status`, `run`) | `migrate` |
 | `commands/serve.py` (~205 lines) | `serve` production foreground server with `--workers` option. Metabase setup failure is non-fatal (prints warning, continues to uvicorn). | `serve()` |
@@ -102,7 +103,7 @@ dango (top-level group)
 ├── dashboard (group)           ← commands/dashboard.py
 │   ├── provision
 ├── mcp (group)                 ← commands/mcp_server.py
-│   ├── run                     ← commands/mcp_server.py (FastMCP stdio server)
+│   ├── run                     ← commands/mcp_server.py (FastMCP stdio server; helpers in commands/mcp_helpers.py)
 │   └── setup, status           ← commands/mcp_setup.py
 ├── migrate (group)             ← commands/migrate.py
 │   ├── status, run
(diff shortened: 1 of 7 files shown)


aaronteoh:
## Review: complete and correct, recommend merge

Author's own review after three rounds — implementation, independent review round 1, independent review round 2 (scoped to the round-1 diff) — plus a mid-session rebase onto `v1.0.8` after Session D merged.

**What this PR does:** adds `dango mcp` (`run`/`setup`/`status`): a local-stdio-only fastmcp server exposing 8 read-only tools, plus LLM-client config auto-detection/writing.

**Why I'm confident:**

- **Every fix was verified empirically, not just re-read.** The two highest-risk bugs found in review — `query()`'s timeout not actually cancelling the running DuckDB query, and the atomic-write fix silently downgrading `settings.json` to 0600 — were each confirmed with a standalone repro *before* fixing, and each fix has a positive-control regression test (a connection-spy wrapper proving `conn.interrupt()` fires; a permission-mode assertion) rather than a test that would pass against either the buggy or fixed version.
- **10 + 6 independent-review findings were evaluated individually**, not applied wholesale: 2 rejected as false positives with reasoning (different threat model for raw error text; row-limit cap is spec-mandated), 2 accepted as documented trade-offs/out-of-scope (no audit logging — no authenticated-user concept to log against; a duplicate atomic-write pattern in an unrelated file, left alone rather than refactoring something this PR doesn't otherwise touch), the rest fixed.
- **The 4 original spec-pseudocode bugs were structural, not cosmetic** (a phantom function, a broken `app.state.project_root` dependency chain that would have silently degraded every source to `status: unknown`, a SELECT-only check that both rejected valid CTEs and didn't block multi-statement injection) — catching these before merge was the point of the pre-implementation grep pass.
- **Full test suite green throughout**: `pytest -x` foreground, run after every structural change (async conversion, both file splits, the interrupt-on-timeout rewrite, the rebase) — last run 4339 passed, 0 failed. `ruff`/`mypy` clean on every touched file. CI green across all 9 checks.
- **Real protocol-level verification, not just unit tests**: spawned `dango mcp run` as an actual subprocess and drove it with `fastmcp.Client` over real stdio transport after every major change, confirming no JSON-RPC corruption and that tools behave correctly end-to-end (not just when called directly in-process).
- **`dango mcp setup`/`status` exercised against isolated `HOME=` overrides**, never the real environment, confirming both the LLM-client-detection and config-write paths work correctly including the fixed venv-console-script resolution and permission preservation.
- **Rebase onto `v1.0.8` post-Session-D-merge was clean** (the only overlap was the two independent, non-conflicting lines in `main.py` the task spec predicted), re-verified in full afterward, not assumed.

**Known, deliberate non-fixes** (documented in the PR body, not oversights): no audit logging on `query()` (architectural mismatch, no user to attribute reads to), row limit hardcoded at 500 (spec-mandated), the duplicate atomic-write pattern in `dango/governance/pii_overrides.py` (pre-existing, unrelated file, flagged as a follow-up candidate).

No exemptions or skip-mechanisms used anywhere in this PR — every check that needed to pass, passed by fixing the underlying issue (including two real file-size-check splits along the codebase's own established `remote.py` cross-file-registration pattern).

Recommend merging as-is.

PR #453, opened 2026-09-02
[throwaway] path-filter isolation check for #452

aaronteoh:
Throwaway PR to test metabase-integration-test.yml path filter in isolation. Not for merge — will close after verifying. See PR #452.


aaronteoh:
Confirmed the path filter isolation result — metabase-integration-test.yml did not run for this PR (verified via both gh pr checks and gh run list --workflow=metabase-integration-test.yml, which returned zero runs). Closing this throwaway verification PR without merging.

PR #478, opened 2026-09-07
1.0.8: MCP config redesign — client-native install, project scoping, version safety

aaronteoh:
## Summary
- Replace `dango mcp setup`'s global config-file writing with the client's own native install command where one exists (verified live for Claude Code: `claude mcp add --scope local`), eliminating cross-project config bleed and matching how established MCP servers (GitHub's, Sentry's) handle multi-client installation.
- Cursor gets its own fix, not a fallback: a project-scoped, git-committable `.cursor/mcp.json` using a bare `dango` command + Cursor's `${workspaceFolder}` substitution (confirmed live via docs) — this eliminates both the cross-project-bleed bug *and* the version-pinning bug for Cursor specifically.
- Windsurf confirmed still global-only (checked its 2026 docs/changelog) — `DANGO_PROJECT_ROOT` is now injected into the global entry, and both the CLI output and docs state the one-project-at-a-time limitation explicitly.
- Add `DANGO_PROJECT_ROOT` as the primary project-resolution signal for MCP tools (`mcp_helpers._get_project_root()`), falling back to cwd-walking only when unset — matching the existing convention already used in `serve.py`, `scripts_api.py`, and `platform/scheduling/jobs.py`.
- Add a version-compatibility warning (`_check_version_compatibility()`, stderr-only) printed once at MCP server startup when the running dango differs from a project's own recorded version.
- Add `dango mcp remove` (previously no teardown command existed at all), reusing a new shared `_atomic_write_json()` helper factored out of `_write_mcp_config()` rather than duplicating the atomic-write logic.
- Companion docs PR: getdango/docs#52.

## Phase 1 investigation findings (all live-verified, not assumed)

**1a. `claude mcp add --scope local` — works, but the prompt's assumed arg order is wrong.**
`--env` is variadic and greedily swallows the next token, so `claude mcp add --scope local --env KEY=VAL name -- cmd` actually fails:
```
$ claude mcp add --scope local --env DANGO_PROJECT_ROOT=$(pwd) dango-test -- /bin/echo test
Invalid environment variable format: dango-test, environment variables should be added as: -e KEY1=value1 -e KEY2=value2
```
Working form places `--env` after the name:
```
$ claude mcp add --scope local dango-test --env DANGO_PROJECT_ROOT=$(pwd) -- /bin/echo test
Added stdio MCP server dango-test with command: /bin/echo test to local config
File modified: /Users/aaronteoh/.claude.json [project: /private/tmp/.../mcp-test-a]
```
Confirmed it writes to `~/.claude.json` under `projects["<abs-path>"].mcpServers`, confirmed invisible from a different directory (`claude mcp list` from a sibling dir shows nothing), confirmed `claude mcp get <name>` reports `Scope: Local config (private to you in this project)` and even prints the exact removal command, and confirmed `claude mcp remove <name> --scope local` fully clears the entry.

**1b. Cursor has no native CLI, but `${workspaceFolder}` substitution is real** (confirmed via two independent doc checks) — this makes a portable, git-committable `.cursor/mcp.json` viable, so Cursor gets the good fix, not the global-fallback hedge the prompt anticipated.

**1c. Windsurf confirmed still global-only as of 2026-09-08** (checked its changelog/docs) — no project scoping exists on the client side.

**1d. FastMCP roots — live end-to-end test, not just a docs read: fails under the actual client/server pairing dango ships.** Built a throwaway FastMCP server with a `get_roots` tool calling `ctx.session.list_roots()`, registered it via `claude mcp add --scope local`, invoked it through a real non-interactive Claude Code session:
```
$ claude -p "Call the mcp__roots-test__get_roots tool ..." --allowedTools "mcp__roots-test__get_roots"
{"result":"ERROR: NoBackChannelError: Cannot send 'roots/list': this transport context has no back-channel for server-initiated requests."}
```
Root cause (confirmed reading installed `fastmcp`/`mcp` source): the MCP protocol version Claude Code negotiates by default (`SEP-2577`, "2026-07-28 era") removed the server-initiated-request back-channel `roots/list` depends on. Per the prompt's own instruction for this outcome: skipped entirely, not stubbed out.

## Verification
- `pytest -x`: 4549 passed, 30 skipped, 0 failed (full suite, twice — before and after a test-file split for the 500-line file-size check).
- `ruff check dango/`: passed. `ruff format --check`: passed. `mypy dango/`: passed (240 source files).
- Live-verified end-to-end against a real scratch project (not `tests/beta-1`): `dango mcp setup` → wrote the correct local-scope Claude Code entry with `DANGO_PROJECT_ROOT` set → `claude mcp list` from a different directory shows nothing → `dango mcp status` reports configured from the project directory → `dango mcp remove` → `~/.claude.json` entry confirmed empty → `dango mcp status` confirms removed.
- Live-verified the version-mismatch warning fires against the real `_get_project_root()`/`_check_version_compatibility()` functions with a real project (`dango_version: "1.0.7"` recorded, running version monkeypatched to `1.0.8`): produced the exact expected warning string.
- All scratch MCP entries (`~/.claude.json`) and scratch directories cleaned up after every live test — verified empty before finishing.

## Regression check
- A project with `DANGO_PROJECT_ROOT` unset still resolves via cwd exactly as before (`test_falls_back_to_cwd_when_env_var_unset`), and this is also true for every existing MCP tool caller (`mcp_server.py` x8, `mcp_mutations.py` x6) — none pass any args to `_get_project_root()`, so the change is a safe drop-in.
- `dango mcp setup`'s "no LLM clients detected" fallback message still works unchanged.
- `_write_mcp_config()` is preserved (now shared by Cursor and Windsurf, plus a new `_atomic_write_json()` helper factored out for `_remove_mcp_config()` to reuse) — Windsurf's status-check JSON-file-reading logic keeps working.
- Deliberate behavior change, called out explicitly: `dango mcp setup`/`status`/`remove` now require running inside a Dango project (`require_project_context`), since Cursor's config is now project-scoped and Claude Code's local-scope lookup is inherently per-project — there's no longer a "check global client config with no project context" mode.

## Test plan
- [x] `pytest -x` — 4549 passed, 0 failed
- [x] `ruff check` / `ruff format --check` / `mypy` — all clean
- [x] Live end-to-end verification against a real scratch project (see Verification above)
- [x] Scratch state cleaned up (`~/.claude.json` verified empty for all test project keys)

https://claude.ai/code/session_01MyYKvCUZcrvQwMHh1sNWEp


aaronteoh:
## Independent code review (post-merge-approval, different pass than the implementation)

Reviewed the full diff fresh (not from memory) and live-tested two behavioral edge cases the original implementation and its unit tests didn't cover: **re-running `dango mcp setup` when already configured**, and **`dango mcp remove` when there's nothing to remove**. Found and fixed 3 real bugs, pushed as `d609fab`. CI will need to re-run on this new commit before merge.

### Fixed now (3)

**1. `dango mcp setup` was not idempotent — every re-run after the first would fail for Claude Code.**
`claude mcp add` has no `--force`/overwrite flag. Live-confirmed:
```
$ claude mcp add --scope local dango-review --env ... -- /bin/echo test
Added stdio MCP server dango-review ...
$ claude mcp add --scope local dango-review --env ... -- /bin/echo test
MCP server dango-review already exists in local config   # exit 1
```
This meant the *exact* scenario BUGS-FOUND.md's version-pinning entry exists to fix — re-running `dango mcp setup` after switching venvs to update the pinned path — would instead report a spurious failure and drop Claude Code entirely from `configured` (which, for a Claude-Code-only machine, would incorrectly print "No LLM clients detected"). Fixed by removing any existing local-scope entry first (ignoring that call's result — it fails harmlessly with nothing to remove on a first run) before adding. Live-verified 3 consecutive `dango mcp setup` runs against a real scratch project, all clean.

**2. `dango mcp remove` printed an alarming warning for Claude Code on the normal "nothing to remove" case.**
`claude mcp remove` on a nonexistent entry exits 1 with `No MCP server named "dango" in local scope` (live-confirmed). The code treated any nonzero exit as worth a yellow warning, so a project using only Cursor (no Claude Code ever configured), or a second `dango mcp remove` run, would print a scary-looking warning for a no-op — inconsistent with Cursor/Windsurf's silent handling of the identical case via `_remove_mcp_config()`. Fixed by detecting the "No MCP server" message and treating it as silent, matching Cursor/Windsurf. A genuine failure (any other message) still warns — covered by a new test.

**3. No `timeout=` on any of the three `claude mcp` subprocess calls.**
Checked the codebase's own convention (`platform/docker.py`) for external-CLI subprocess calls: all of them set a `timeout=` and catch `except (subprocess.TimeoutExpired, FileNotFoundError)`. None of the three new `claude mcp add/get/remove` calls had this — a hung `claude` process would hang `dango mcp setup`/`status`/`remove` indefinitely with no escape. Added `timeout=10` and widened the exception handling to match the house pattern.

All three: live-verified against the real `claude` CLI and a real scratch project (not just mocked), plus new/updated unit tests (54 → still 54 in the setup/server files, +3 net after a further file split to stay under the 500-line hard limit — `_resolve_dango_cmd`/`_write_mcp_config`/`_remove_mcp_config` tests moved to new `tests/unit/test_mcp_setup_helpers.py`). Full suite: 4552 passed, 0 failed. `ruff`/`ruff format`/`mypy` all clean.

### Accepted trade-offs (2)

- **`mcp_status`/`mcp_remove` parse `claude mcp get/remove`'s human-readable stdout/stderr text** (`"Local config" in result.stdout`, `"No MCP server" in message`) rather than structured output. *Reason: `claude mcp get`/`list`/`remove` have no `--json` or equivalent structured-output flag (checked `--help`) — string-matching is the only mechanism available today. Isolated to one conditional each; trivial to swap if Claude Code adds structured output later.*
- **Cursor's code path (`_setup_cursor`, the project-scoped `.cursor/mcp.json` + `${workspaceFolder}` write) was never live-tested against a real Cursor installation** — only unit-tested with mocks. *Reason: Cursor isn't installed on this dev machine (no app, no CLI, no `~/.cursor`). The `${workspaceFolder}` substitution claim is corroborated by two independent doc lookups but is genuinely unverified against real Cursor behavior. Flagging as residual risk rather than silently treating it as equally solid as the Claude Code path, which *was* live-verified end-to-end.*

### False positives considered and dismissed

- *Relative `DANGO_PROJECT_ROOT` values could resolve against the wrong cwd.* Doesn't apply in practice — every path dango itself ever writes into a config entry is already absolute (`find_project_root()`'s return value, or Cursor's own `${workspaceFolder}` resolution happens client-side before the env var reaches the process). Only a manually hand-edited config could hit this, which is a pre-existing, unrelated user-error case.
- *`mcp_status` doesn't verify the Claude Code entry's actual command/args, only that an entry named `dango` exists at local scope.* Not a regression — matches the original code's depth of verification (checked for `"dango" in servers`, nothing deeper).

### One process note, not a code finding
Early in this session, a pre-commit hook failure on `git commit` left `HEAD` unexpectedly at the inherited branch tip (the `v1.0.8` merge commit), and a subsequent `git commit --amend --no-edit` silently absorbed my changes into *that* commit with its original message instead of creating a new one. Caught immediately via `git log`/`git diff` before pushing, recovered with `git reset --soft` to the correct parent, verified the diff was exactly my intended change set, and recommitted properly — no bad state was ever pushed. Mentioning it because it's a sharp edge worth remembering: a failed pre-commit hook means no new commit was made, so `--amend` targets whatever was already at `HEAD`, not "the commit you just tried to make."

PR #479, opened 2026-09-07
1.0.8: Housekeeping fixes — stale count, status port, telemetry disclosure, duplicate warning

aaronteoh:
## Summary
- Fix 4 independent, low-risk findings from BUGS-FOUND.md (manual review, 2026-09-06/07):
  - `dango source add --help`'s stale "27+ sources" count (4th location this cycle, PR #463 missed it)
  - `dango status`'s hardcoded "Metabase (port 3000)" label (3rd instance of a bug class fixed twice
    already this cycle, PRs #459/#460)
  - Telemetry settings page/CLI never disclosed dango/dbt/dlt are machine-wide, Metabase is
    project-scoped; `dango init`'s consent-skip path was completely silent on every project after
    the first
  - `dango model add`/`source add` printed two separate, redundant git-branch warnings on main/master

## Verification
- `pytest -x`: 4540 passed, 30 skipped, 0 failed
- `ruff check` / `ruff format --check` / `mypy dango/`: all clean on changed files
- `make css-build`: zero diff on `tailwind.min.css` (telemetry.html change only used existing
  utility classes, as expected)
- Live-verified `dango status`'s fix by running the real `status()` command end-to-end (all
  external deps mocked, no exception swallowed) with `platform.metabase_port=3001` — output shows
  `Metabase (port 3001)`, confirmed absent `Metabase (port 3000)`
- Live-verified `dango model add`/`source add` on real git repos (real `git init`, branch `main`)
  through the actual wizard code path up to the warning print (only the interactive prompt step
  mocked) — output contains the plain `Warning:` line exactly once, never the old boxed
  "Git Branch Reminder" panel
- Live-verified `dango init`'s second-project informational line and that the CI/opt-out skip
  paths stay fully silent, via `_prompt_telemetry_consent()` unit tests

## Regression check
- `check_git_branch_warning()` still has other live callers (`source_edit`... actually
  `source remove`/`sync` at `dango/cli/commands/source.py` lines ~446, ~762) — untouched
- `mcp_helpers.py::_git_warnings()` — untouched, separate MCP call site
- `tests/unit/test_cli_sync.py` — no delta (24 tests, all pass)
- `tests/unit/test_egress_allowlist.py`, `test_telemetry.py` — no delta beyond the new tests in
  `test_cli_init_telemetry.py`

https://claude.ai/code/session_01MyYKvCUZcrvQwMHh1sNWEp

### DIFF
--- a/dango/cli/commands/model.py
+++ b/dango/cli/commands/model.py
@@ -39,13 +39,10 @@ def model_add(ctx: click.Context) -> None:
       dango model add    Run interactive wizard
     """
     from ..model_wizard import add_model
-    from ..utils import check_git_branch_warning, require_project_context
+    from ..utils import require_project_context
 
     try:
         project_root = require_project_context(ctx)
-
-        # Check git branch (gentle reminder if on main/master)
-        check_git_branch_warning(project_root)
 
         model_path = add_model(project_root)
 
--- a/dango/cli/commands/platform.py
+++ b/dango/cli/commands/platform.py
@@ -1023,9 +1023,11 @@ def status(ctx: click.Context) -> None:
                 status_text = (
                     f"[{svc_status.value}]● {svc_status.value.capitalize()}[/{svc_status.value}]"
                 )
-            table.add_row("Metabase (port 3000)", status_text)
+            table.add_row(f"Metabase (port {config.platform.metabase_port})", status_text)
         else:
-            table.add_row("Metabase (port 3000)", "[red]● Stopped[/red]")
+            table.add_row(
+                f"Metabase (port {config.platform.metabase_port})", "[red]● Stopped[/red]"
+            )
 
         console.print(table)
         console.print()
--- a/dango/cli/commands/source.py
+++ b/dango/cli/commands/source.py
@@ -87,20 +87,13 @@ def source_add(ctx: click.Context) -> None:
     """
     Add a new data source (interactive wizard).
 
-    Supports 27+ sources across 9 categories:
-      - Marketing & Analytics (7): Facebook Ads, Google Ads, Sheets, Analytics, etc.
-      - Business & CRM (7): HubSpot, Salesforce, Zendesk, Jira, etc.
-      - E-commerce & Payment (1): Stripe
-      - Files & Storage (2): Notion, Email Inbox
-      - Databases (1): MongoDB
-      - Streaming (2): Kafka, Kinesis
-      - Development (1): GitHub
-      - Communication (1): Slack
-      - Local & Custom (2): CSV, REST API
+    Supports 35+ data source types across marketing/analytics, business/CRM,
+    e-commerce, files/storage, databases, streaming, development, and
+    communication tools. Run this command to see the full current list.
     """
     from dango.cli.source_wizard import add_source
 
-    from ..utils import check_git_branch_warning, check_v01x_project
+    from ..utils import check_v01x_project
 
     check_v01x_project()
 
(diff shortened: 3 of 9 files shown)


aaronteoh:
## Independent self-review

Reviewed this PR with fresh eyes after opening it (manual pass + `/code-review` skill run against the PR diff). Two findings surfaced; categorized below.

### Fix now (fixed in `9e09ea4`)

**`dango/cli/init.py` — `_prompt_telemetry_consent()` swallowed the disclosure for inherited opt-outs.**

The original gate order was:
```python
if is_ci() or not is_telemetry_enabled():
    return
if has_recorded_consent():
    console.print(...)  # "already configured machine-wide (on/off)"
    return
```
`is_telemetry_enabled()` itself returns `False` whenever a prior "no" answer is on record (it's one of the opt-out signals it checks, alongside `DO_NOT_TRACK`/`DANGO_TELEMETRY`/config). So when a user opted out in project A, `dango init` in project B hit `not is_telemetry_enabled()` and returned *before ever reaching* `has_recorded_consent()` — the new informational line only ever fired for the opted-*in* case. That's the opposite of what this fix was for: silently inheriting a decision from another project with zero disclosure, in exactly the "off" scenario a user is most likely to be surprised by.

Reproduced live before the fix:
```
has_recorded_consent: True
is_telemetry_enabled: False
--- dango init in project B (should print, was silent) ---
(nothing printed)
```
And confirmed fixed after reordering to check `has_recorded_consent()` before the `is_telemetry_enabled()` gate:
```
Telemetry: already configured machine-wide (off) — `dango telemetry status` to
review, `dango telemetry off --all` to change.
```
CI and "opt-out with no prior recorded answer" (pure `DO_NOT_TRACK`/config, never having answered the prompt) still stay fully silent — verified via `test_ci_skip_still_silent` and `test_opted_out_skip_still_silent`, both still passing. Added a new regression test, `test_second_init_prints_info_line_when_consent_previously_declined`, that pins this exact scenario.

### Accepted trade-off

**`dango source add --help`'s "35+ data source types" doesn't match what the wizard's picker actually lists.**

Verified: `SOURCE_REGISTRY` has 34 entries, but only 26 have `wizard_enabled=True` — that's the actual number of options `_select_source_flat()` shows when you run the wizard. The "35+" I wrote comes from `SourceType` enum count, which is the config *schema's* total (types you can hand-write into `sources.yml`), not the wizard's interactive list.

Not changing this, because:
- The task spec explicitly prescribed this exact number and this exact verification command (`len(list(SourceType))`).
- "35" is the project's own already-established canonical figure — it's what `README.md`, the docs ingestion table, and the roadmap positioning line all say (`"35 data sources"`, fixed by PR #463). Citing 26 here would create a *fourth* inconsistent number, not fewer.
- 26 (`wizard_enabled` count) is no more future-proof than 35 — it's just as capable of going stale, for a narrower and less-recognizable concept.

Related pre-existing (not introduced by this PR) inconsistency worth tracking separately: `dango/ingestion/sources/registry.py` and `dango/ingestion/CLAUDE.md` still say "33 sources", and `dango/cli/source_wizard.py:3`'s docstring still says "27+" — three different numbers for related-but-distinct concepts, none of which this housekeeping PR's scope covers (it was explicitly scoped to only `source.py`'s docstring). Flagging this as a new finding for `BUGS-FOUND.md` rather than silently dropping it or unilaterally expanding this PR's scope to fix it.

### False positives
None — both findings from the automated pass held up under verification.

---
**Verification of the fix-now item**: `pytest -x` — 4541 passed, 30 skipped, 0 failed (was 4540/30 before this fix — the +1 is the new regression test). `mypy dango/cli/init.py` — clean. `ruff check`/`ruff format --check` — clean.

PR #482, opened 2026-09-08
1.0.8: dango status shows running MCP server processes

aaronteoh:
## Summary
- `dango status` now detects and reports a running MCP server process for the current project,
  matched via the DANGO_PROJECT_ROOT env var set by `dango mcp setup` (PR #478) — read-only
  detection, Dango does not manage the MCP process's lifecycle.
- Added `_find_mcp_server_process()` in `dango/cli/commands/platform.py`, mirroring
  `kill_orphan_watchers()`'s `psutil.process_iter()` scan shape (in
  `dango/platform/local/watcher_lifecycle.py`) rather than shelling out to `pgrep`/`ps`: neither
  tool is used anywhere else in this codebase, and `process_iter()` is the existing tested pattern
  for "scan all processes by cmdline, verify identity via environ()."

## Deviation from original task spec
The task prompt's sample code used `subprocess.run(["pgrep", "-f", "dango mcp run"])`. Verified live
that `pgrep` is unused anywhere in this codebase and that `psutil` is not currently imported in
`platform.py` at all (both claims in the original prompt were wrong, confirmed with the coordinating
session). Found a closer existing precedent instead — `kill_orphan_watchers()` — and mirrored its
`psutil.process_iter()` shape: no new external-tool dependency, more directly testable (mock
`psutil.process_iter` the same way `test_watcher_lifecycle.py` already does).

## Verification
- `pytest -x`: 4572 pass, 30 skip, 0 fail (full suite, not just the new test file)
- `ruff format --check dango/`: 240 files already formatted
- `mypy dango/`: Success, no issues found in 240 source files
- **Live-verified** (two throwaway `dango init --skip-wizard` scratch projects, not `tests/beta-1/`):
  started a real `dango mcp run` process with `DANGO_PROJECT_ROOT` pointed at scratch project A
  (kept its stdin open indefinitely via `tail -f /dev/null | dango mcp run`, since the stdio server
  exits on stdin EOF). `dango status` from project A showed:
  `MCP server │ ● Running (PID 93190)` (green). `dango status` from project B (a different
  initialized project, same live MCP process still running) showed:
  `MCP server │ ● Not running` (dim) — proving the DANGO_PROJECT_ROOT match actually discriminates
  by project, not "any dango mcp run process counts." Killed the test process and removed both
  scratch projects before finishing.
- **Bug caught by the live test, not the unit tests**: a `dango` console script has a
  `#!/path/to/python` shebang, so the kernel rewrites the process's actual argv to
  `[<python-interpreter>, <path>/dango, "mcp", "run"]` — `"dango"` is NOT `argv[0]` as the original
  implementation assumed. Fixed by scanning the whole cmdline list for a `dango`/`dango.exe`
  basename instead of assuming a fixed index, and added a regression test
  (`test_find_mcp_server_process_detects_shebang_rewritten_argv`) using the exact argv shape
  observed via `ps aux` against the real spawned process.

## Regression check
- Confirmed this never starts/stops/signals the MCP process it detects — reviewed the full diff,
  no `terminate`/`kill`/`send_signal`/`Popen` calls anywhere in the new code
- Confirmed existing Web UI/Watcher/Metabase rows in `dango status` are unchanged
- Not-running MCP server uses `[dim]● Not running[/dim]`, not `[red]● Stopped[/red]` — verified in
  both the unit test (`test_status_shows_mcp_server_not_running` asserts `"Stopped" not in output`)
  and the live test output above

### DIFF
--- a/dango/cli/commands/platform.py
+++ b/dango/cli/commands/platform.py
@@ -3,6 +3,8 @@
 Platform lifecycle commands (start, stop, status) and port helpers.
 """
 
+from pathlib import Path
+
 import click
 
 from dango.cli import console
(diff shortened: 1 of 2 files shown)


aaronteoh:
## Independent code review (post-implementation, different pass than the one that wrote it)

Re-read the full diff fresh, checked it against the argv/env assumptions live-verified earlier,
and looked for correctness bugs, coverage gaps, and reuse/simplification opportunities.

### Fixed now (pushed as `c783232`)
- **Test coverage gap**: existing tests covered "cmdline matches but `DANGO_PROJECT_ROOT` points at
  a *different* project" but not "cmdline matches and `DANGO_PROJECT_ROOT` is *absent entirely* from
  `environ()`" — the exact pre-1.0.8-OPS-4 / hand-configured case the task spec explicitly calls out
  ("some MCP processes may predate this env var... don't guess, just don't count them as this
  project's"). Added `test_find_mcp_server_process_ignores_process_missing_env_var`. Code path was
  already correct (`if env_root and ...` — falsy on a missing key), only the test was missing.

### Accepted trade-offs (no change)
- **Cmdline matching is a loose heuristic** (`"dango"/"dango.exe"` basename anywhere in argv, plus
  `"mcp"` and `"run"` as separate tokens) — theoretically a third-party tool with an unrelated `mcp
  run` subcommand invoked from a path containing a `dango`-named script could false-positive. Same
  risk class and same looseness as the existing `kill_orphan_watchers()` precedent this mirrors
  (substring match on `"watcher_runner.py"` + resolved root), so it matches this codebase's existing
  risk tolerance for this style of check rather than introducing a new one.
- **`python -m dango.cli.main mcp run` invocation would not be detected** (no argv token has basename
  `"dango"` in that form) — not a real gap: `mcp_setup.py`'s `_resolve_dango_cmd()` always spawns the
  resolved console-script path (or bare `"dango"` on PATH), never `python -m`, for every real client
  (Claude Code, Cursor, Windsurf). The `-m` form is only used in this repo's own worktree-editable-install
  smoke-testing convention, never by a real MCP client.
- **No explicit timeout/cap on the `psutil.process_iter()` scan** — matches `kill_orphan_watchers()`
  exactly; these are local `psutil` calls (no subprocess, no network), and the original prompt's
  own regression-risk section already reasoned through this and accepted it.
- **Windows behavior unverified** — `dango.exe` naming is handled in the match, but there's no
  Windows CI job or live Windows environment to test the shebang-equivalent argv shape for
  pip-generated `.exe` launchers. Consistent with the rest of this codebase's Windows coverage gap
  (no special-cased Windows argv handling exists anywhere else either); not blocking for this PR.

### False positives (checked, don't apply)
- **Possible TOCTOU between the cmdline check and `proc.environ()` call** (process exits in between)
  — already covered: `NoSuchProcess`/`ZombieProcess` are both caught and the loop continues to the
  next candidate.
- **Redundant `int(proc.info["pid"])` cast** — not redundant: `proc.info` is typed loosely by
  psutil's stubs, so the explicit cast is what lets `mypy` verify the `int | None` return
  contract; removing it would very likely reintroduce a mypy error.
- Re-confirmed the `terminate`/`kill`/`send_signal` assertions in
  `test_find_mcp_server_process_never_kills_or_signals` are genuine positive-control-style checks
  against a real `MagicMock`, not a no-op/tautological test — matches this project's
  positive-control-verification convention.

### Overall
No blocking issues. One real test-coverage gap found and closed. Everything else is either already
handled correctly or an accepted trade-off consistent with existing codebase precedent, not a new
risk introduced by this PR. Full suite still green after the added test (`pytest -x`: 4573 pass, 30
skip, 0 fail locally; CI pending at time of this review).

PR #483, opened 2026-09-08
1.0.8: Fix broken MySQL source type (SourceType/registry drift)

aaronteoh:
## Summary
- Fix a real, live-confirmed bug: `dango source add` crashed with a Pydantic ValidationError when a
  user selected MySQL and completed the wizard, losing everything they entered. Root cause: "mysql"
  existed in SOURCE_REGISTRY with wizard_enabled: True but was never added to the SourceType enum.
- Added SourceType.MYSQL = "mysql" (Databases group, alongside POSTGRESQL, which uses the identical
  dlt `sql_database` mechanism and works fine today).
- Removed SourceType.SCRAPY = "scrapy": grep confirmed zero references anywhere in the codebase
  outside its own enum definition (`grep -rn "scrapy\|SCRAPY" dango/ --include="*.py"` → only the
  enum line itself and a `registry.py` comment that explicitly said "scrapy not available"; the
  same grep against `tests/` returned zero hits). No registry entry, no `dlt_runner.py` handling,
  no wizard entry — vestigial, same class of drift as the mysql bug but lower severity since nobody
  could ever select it.
- Fixed 3 stale source-count mentions (`registry.py`, `source_wizard.py`, `ingestion/CLAUDE.md`) from
  the old "33"/"27+" counts to the real, live-verified count of 34
  (`len(SOURCE_REGISTRY)` = 34; `len(list(SourceType))` = 35 = 34 registry entries + `sql_database`,
  which is intentionally enum-only with no registry entry).

`sql_database` is untouched — confirmed via `git diff` that `registry.py`'s only changes are
comment/docstring updates (stale counts + the now-obsolete "scrapy not available" comment); no
registry entries were added or removed.

## Verification
- `pytest -x` (full suite, not a subset): **4564 passed, 30 skipped, 0 failed** (~239s)
- `ruff format --check dango/`: 240 files already formatted
- `mypy dango/`: Success, no issues in 240 source files
- Live-verified: `DataSource(name="t", type="mysql")` no longer raises (confirmed inside the fixed
  worktree; a first attempt without `PYTHONPATH` pointing at the worktree hit the main clone's stale
  editable install and reproduced the exact pre-fix error — itself further live confirmation of the
  bug, and a reminder of the known worktree editable-install gotcha).
- Regression test confirmed to fail against the pre-fix code, pass against the fix: temporarily
  stashed the `models.py` enum change and re-ran `TestSourceTypeRegistryDrift` — all 3 new tests
  failed with the exact bug signature (`wizard_enabled registry entries with no matching SourceType
  enum value: ['mysql']`, and the scrapy `ValueError`-not-raised failure). Restored the fix and all 3
  passed again.
- Live end-to-end (with one disclosed substitution): true interactive-TTY arrow-key automation via
  `expect` was not reliably drivable in this sandbox — there's no controlling tty for `stty` to size
  the pty, and python-inquirer's raw-mode key reads did not consume queued arrow-key escape sequences
  sent over the pty (they echoed back unconsumed instead of navigating the list). Instead, drove the
  actual `SourceWizard` class — the exact object `dango source add` instantiates via
  `dango.cli.source_wizard.add_source()` — against a real scratch project, with only `inquirer.prompt`
  answered programmatically. Every other step ran as real, unmocked production code: the parameter
  collection state machine, `DataSource` construction, Pydantic validation, the `sources.yml` write,
  and a `ConfigLoader.validate_config()` + reload round-trip. Result: wizard selected MySQL, named the
  source, reused a pre-populated `.env` credential, saved successfully, config validated, and the
  reloaded `DataSource.type` came back as `SourceType.MYSQL`.

## Regression check
- Confirmed no exhaustive match/case over SourceType exists that the new/removed enum members could
  break: `grep -rn "match source_type\|match .*\.type:\|match.*SourceType" dango/` → zero hits,
  re-verified against current `v1.0.8` code (not just the original investigation's claim).
- `get_source_metadata()` unaffected — plain dict `.get()` lookup by registry key string, unchanged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -260,7 +260,7 @@ dango/                          # Python package source
 │   ├── dlt_runner.py           # ⚠ 2885 lines — orchestrates full sync pipeline
 │   ├── csv_loader.py           # Multi-format file loading with dedup (922 lines)
 │   ├── sources/
-│   │   └── registry.py         # Source metadata (33 source types)
+│   │   └── registry.py         # Source metadata (34 source types)
 │   └── dlt_sources/            # 127 files — helper files are custom Dango code (safe to modify); vendored connectors should not be changed
 │
 ├── transformation/             # Level 1 — dbt model generation & execution
--- a/dango/cli/source_wizard.py
+++ b/dango/cli/source_wizard.py
@@ -1,6 +1,6 @@
 """dango/cli/source_wizard.py
 
-Metadata-driven wizard that works for all 27+ data sources. Uses SOURCE_REGISTRY for display names, categories, and parameters.
+Metadata-driven wizard that works for all 34 data sources. Uses SOURCE_REGISTRY for display names, categories, and parameters.
 """
 
 from pathlib import Path
--- a/dango/config/CLAUDE.md
+++ b/dango/config/CLAUDE.md
@@ -69,3 +69,5 @@ Loads, validates, and manages dango project configuration files (project.yml, so
 | `models.py` field names | Changing field names breaks existing project.yml/sources.yml files |
 | `models.py` `SourceType` enum values | Enum values are stored in user config files and referenced across modules |
 | `exceptions.py` re-export shim | Exception classes are caught by name in cli/, web/, and ingestion/. Adding/removing classes here must mirror `dango/exceptions.py`. |
+
+Carve-out for `SourceType`: a member with zero registry entry, no wizard path, and no `dlt_runner.py` handling (confirmed via grep) was never reachable and can't exist in a real persisted config, so it can be removed — the rule protects values a live config could actually contain, not dead ones (see 1.0.8-Q3's removal of `SCRAPY`).
--- a/dango/config/models.py
+++ b/dango/config/models.py
@@ -21,7 +21,7 @@ class DeduplicationStrategy(str, Enum):
 
 
 class SourceType(str, Enum):
-    """Data source types — 35 source types (27 dlt verified + CSV + Local Files + REST API + dlt_native + Filesystem + PostgreSQL + sql_database + Scrapy)"""
+    """Data source types — 35 source types (27 dlt verified + CSV + Local Files + REST API + dlt_native + Filesystem + PostgreSQL + MySQL + sql_database)"""
 
     # Local/Custom
     CSV = "csv"
@@ -57,9 +57,10 @@ class SourceType(str, Enum):
     NOTION = "notion"
     INBOX = "inbox"
 
-    # Databases (3)
+    # Databases (4)
     MONGODB = "mongodb"
     POSTGRESQL = "postgres"
+    MYSQL = "mysql"
     SQL_DATABASE = "sql_database"  # Generic for 24 SQL databases via dlt
 
     # Streaming (2)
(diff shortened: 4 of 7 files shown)


aaronteoh:
## Independent code review (different pass — reviewer hat, not the author hat)

Re-read the diff fresh, re-verified the core claims, and grepped for angles the implementation pass might have missed. Findings below, categorized.

### Fix now — done in commit 532c1cd
Found 2 more stale "33 source types" mentions that the original task's own "3 locations" list didn't catch, both in files already touched or explicitly in scope:
- `dango/ingestion/CLAUDE.md` line 24 — the Files-table row for `sources/registry.py` (a *different* line from the two already fixed in this same file's Purpose/Source-Selection sections).
- `dango/CLAUDE.md` (repo root) — the repository-structure ASCII diagram: `└── registry.py # Source metadata (33 source types)`.

Both now say 34. Pushed as a follow-up commit rather than amending, per repo convention. CI re-ran on the new HEAD.

### Accepted trade-off
Removing `SourceType.SCRAPY` changes the failure mode for a *hypothetical* pre-existing `type: scrapy` entry in someone's `sources.yml`:
- **Before:** `SourcesConfig` parses fine (scrapy was a valid enum value); the scrapy source only fails at sync time with a clear `"Source type 'scrapy' not found in registry"` from `dlt_runner.py` — every other source in the project keeps working.
- **After this PR:** `sources: list[DataSource]` fails Pydantic validation atomically, so `ConfigLoader.load_sources_config()` raises for the *entire file* — every source in the project would fail to load, not just the scrapy one.

Reason accepted: scrapy was never `wizard_enabled`, never had a registry entry, and has zero references anywhere in `dlt_runner.py` — there is no supported path that could have ever produced a `type: scrapy` entry in a real project's `sources.yml`. The only way to hit this is hand-authoring an undocumented enum value, which was already non-functional. Practical risk is effectively zero, and it's outweighed by closing this exact class of registry/enum drift bug (which is the point of the PR). Flagging for the record, not as a blocker.

### False positive / out of scope (not fixed)
- `registry.py`'s own comment block says `"This registry contains 34 sources in five categories:"` but enumerates 6 numbered categories (1–6). This mismatch **predates this PR** — the diff only touched the total-count number and category-6's text (to add MySQL alongside PostgreSQL), not the "five categories" header wording. Pre-existing doc bug, unrelated to the mysql/scrapy fix; not blocking, could be a trivial separate cleanup.
- No `CHANGELOG.md` entry added. Checked repo convention via `git log` — 1.0.8 changelog entries are batched into separate `docs: fill in CHANGELOG.md [Unreleased] with PRs #N-#M` commits after PRs merge (e.g. the commit immediately preceding this branch's base), not added per-PR. Omission here matches existing practice, not a gap.

### Confirmed correct (re-verified independently, not just re-reading the original claims)
- `git diff` on `registry.py`: zero registry-entry changes, comment/docstring-only — `sql_database` genuinely untouched.
- `grep -rn "scrapy\|SCRAPY" dango/ tests/`: zero references left anywhere outside the (now-removed) enum line.
- `grep -rn "match source_type\|match .*\.type:\|match.*SourceType" dango/`: zero hits — no exhaustive match/case broken by the enum change.
- `sources: list[DataSource]` in `config/models.py` confirmed as the mechanism behind the accepted-trade-off note above (checked directly, not assumed).
- Test suite, ruff, mypy all independently re-confirmed clean on the pre-follow-up-commit state; the two follow-up doc fixes are `.md`-only and don't touch anything the required CI checks (Lint/mypy/Test 3.10/Test 3.12) exercise.

No merge/branch/checkout actions taken — posting this as a comment only.

PR #492, opened 2026-09-11
1.0.8-Z: Rewrite Google OAuth setup instructions

aaronteoh:
## Summary
- Rewrote the Google OAuth setup instructions panel in providers.py to match Google's
  current Auth Platform UI (renamed from "OAuth consent screen", tabbed interface,
  API-before-consent-screen ordering, already-configured-project branch, unverified-app
  warning explanation) — shared by Google Sheets, Google Ads, and Google Analytics
- Follow-up commit: fixed the adjacent "Step 2: Authorize Dango" block, which claimed
  Published apps get a warning-free "standard consent screen" — this contradicted the
  corrected Prerequisites panel above it (per `BUGS-FOUND.md` Gap 3, the unverified-app
  warning depends on scope sensitivity/verification status, not Testing vs. Production).
  Found in independent review of this PR, confirmed against the design doc, fixed in the
  same PR at the coordinator's direction.

## Verification
- `pytest -x`: 4656 pass, 0 fail (30 skipped, pre-existing) — re-run after the Step 2 fix
- `ruff check` / `ruff format --check`: both clean after the Step 2 fix
- Manually ran `dango oauth google_sheets` in a scratch project (via a pty since the
  underlying prompt requires a real terminal), confirmed the panel renders correctly
  with no truncation or broken Rich markup, correct per-source API name substitution
  ("Google Sheets API"), and correct callback URL substitution
- Step 2 fix: rendered the exact updated markup string directly through `rich.Console`
  to confirm no unclosed tags / rendering errors (the full interactive flow stalls at
  the masked Client Secret prompt under a non-controlling pty in this sandbox, so this
  targeted render check was used instead — same markup style as the surrounding
  untouched Testing-mode block, low risk)

## Regression check
- Confirmed via grep this is the only instructions block of its kind (no duplication):
  `grep -rn "Configure the OAuth consent screen\|Google Auth Platform" dango/` returns
  exactly one match, inside `providers.py`
- Text-only change, no functional/API behavior affected
- `ruff format --check` required one auto-format pass on the first commit (double→single
  quote normalization only, no content change) — applied and verified clean

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Nn6RWFt5T69QR6gvQ27igT

PR #493, opened 2026-09-11
1.0.8-Y: Fix OAuth invalid_scope on refresh

aaronteoh:
## Summary
- Fixed invalid_scope RefreshError in source_wizard.py's Google Sheets sheet-fetching
  code — the refresh-only Credentials object was passing a stale/potentially-mismatched
  scope list (the originally *requested* scopes, not what Google actually *granted*)
  into the refresh call, which google-auth's refresh_grant() requires to exactly match.
  Root cause confirmed against the actual installed google-auth library source
  (google/oauth2/_client.py, refresh_grant()): `if scopes: body["scope"] = " ".join(scopes)`.

## Changes
- `dango/cli/source_wizard.py`: removed `scopes=scopes` from the refresh-only
  `Credentials(...)` construction in `_fetch_google_sheets()`, and removed the
  now-dead `scopes = cred.metadata.get("scopes", [])` line (verified unused
  elsewhere in the method).
- Did NOT touch `providers.py`'s `metadata={"scopes": all_scopes}` storage —
  out of scope per the task spec, that data may still be useful for
  display/debugging.

## Verification
- `pytest -x` (foreground): 4657 passed, 30 skipped, 0 failed
- `ruff format --check dango/`: 241 files already formatted
- New unit test `test_google_sheets_refresh_does_not_pass_scopes` added,
  confirming no `scopes` argument reaches the `Credentials` constructor.
  Verified as a genuine regression test via positive control: temporarily
  reverted the fix and confirmed the test fails against the pre-fix code
  (`AssertionError: ... got scopes=[...]`), then confirmed it passes with
  the fix restored.
- **Manual live reproduction: NOT YET DONE.** This requires real Google
  OAuth credentials, which are not available in this session/environment.
  Per the task's acceptance criteria, this fix needs a human to run a live
  end-to-end flow before it can be considered fully verified:
  fresh `dango source add` → Google Sheets → "Set up OAuth now" → complete
  browser consent → enter a real spreadsheet ID → confirm no
  `invalid_scope`/`RefreshError`, in one continuous flow (not the previous
  workaround of aborting and re-running `dango oauth google_sheets`
  separately).

## Regression check
- Confirmed via grep (`cred.metadata.get("scopes"...)` / `metadata["scopes"]`
  patterns across `dango/`) that this is the only call site with this
  pattern in the repo — no other source (Google Ads, Google Analytics, etc.)
  has the identical bug.

## Notes for reviewer
- Test patch target: `Credentials` is imported lazily inside
  `_fetch_google_sheets()` (`from google.oauth2.credentials import
  Credentials`), not at module level in `source_wizard.py`. The task doc
  suggested patching `dango.cli.source_wizard.Credentials`, but that raises
  `AttributeError` (verified) since it's not a module attribute at patch
  time. The test instead patches the actual source module,
  `google.oauth2.credentials.Credentials`.


aaronteoh:
## Independent review of PR #493 (coordinating chat + dispatched subagent, second pass)

**Verified independently:**
- `google.oauth2.credentials.Credentials.__init__`'s `scopes` parameter defaults to `None` (confirmed via `inspect.signature` against the installed venv) — omitting the kwarg is equivalent to `scopes=None`, and `refresh_grant()`'s `if scopes:` check treats `None` as falsy, so `scope` is genuinely dropped from the refresh POST body. The fix does what it claims.
- Confirmed via grep this is the only call site with the `metadata.get("scopes"...)` → refresh-only `Credentials()` pattern in the codebase — no other Google source needs the identical fix.
- CI: Lint, CSS build, egress allowlist, LLM checks, pip-audit, mypy all pass.

**Non-blocking follow-up suggestion:** `providers.py:389` (`metadata={"scopes": all_scopes}`) — the write site for the field this bug was rooted in — has no comment noting the stored value is *pre-exchange requested* scopes, not what Google *actually granted*. A one-line comment there would be cheap insurance against a future Google source reproducing this exact bug. Out of scope for this PR per its task spec; tracked as a suggestion, not a blocker.

**Outstanding before merge:** the manual live-reproduction step (real Google OAuth consent flow) still needs a human with real credentials, as already flagged in the PR description.

Bottom line: fix is correct, minimal, and well-tested. Approved pending the live-reproduction check.

aaronteoh:
## Manual live reproduction — confirmed fixed

Rebased this branch onto latest \`v1.0.8\` (picks up merged #492/#494/#495, no conflicts — only commit touching \`source_wizard.py\` since this branch was cut was this one) so the OAuth instructions text is current too.

Ran a fresh \`dango source add\` → Google Sheets → "Set up OAuth now" → real browser consent (Production-status app, real Google account) → entered a real spreadsheet URL → sheet listing, in one continuous flow, no restart or workaround:

- OAuth exchange succeeded, authenticated as the real account
- Spreadsheet ID/URL accepted (URL form worked, confirming the existing \`docs.google.com\` URL-parsing path)
- **Sheet names fetched successfully — no \`invalid_scope\`, no \`RefreshError\`** — this is the exact code path (\`_fetch_google_sheets()\`'s refresh call right after fresh token exchange) that previously failed
- Source saved, \`dango sync --source test_gsheets\` next step shown

Also incidentally confirmed the unrelated Z fix's warning text is accurate: hit "Google hasn't verified this app" on a Production-status app exactly as the corrected instructions describe (verification status, not publishing status, drives the warning).

All CI green post-rebase. Merging.

PR #495, opened 2026-09-11
1.0.8-W: Set Metabase Site URL for local proxy

aaronteoh:
## Summary
- `setup_metabase()` now sets Metabase's Site URL (`PUT /api/setting/site-url`) to the
  correct `/metabase/`-suffixed, `config.platform.port`-based URL, fixing broken JS asset
  loading (blank page, `Uncaught SyntaxError` on chunk loads) when accessing Metabase
  through the local proxy.
- **Scope expanded per coordinating-chat decision:** `refresh_metabase_connection()` (the
  function that restarts the Metabase container after every sync) also applies the fix,
  since it's the only path that ever reaches a project whose `.dango/metabase.yml` was
  created *before* this fix ships — `setup_metabase()` runs exactly once per project
  (guarded by `credentials_file.exists()`) and cannot retroactively fix already-configured
  projects.
- The actual PUT logic (read `config.platform.port`, build the `/metabase/`-suffixed URL,
  `PUT /api/setting/site-url`, print-and-swallow on failure) is factored into a shared
  private helper `_set_metabase_site_url()`, used by both call sites — no duplication.
- `refresh_metabase_connection()` did not previously authenticate with Metabase's API at
  all (it only restarts the Docker container and polls `/api/health`); it now also logs
  in using the admin credentials already saved in `.dango/metabase.yml` (via the existing
  `_metabase_login()` helper) so it has a session token to call the site-url endpoint with.
  This login + site-url call is wrapped in its own try/except, separate from
  `_set_metabase_site_url()`'s internal one, so any failure (missing/malformed credentials
  file, login failure) never turns a successful container restart into a reported failure.

## Update: independent review caught a real regression, now fixed
An independent code review of this PR (run via the `code-review` skill, cross-checked by
the coordinating chat) found that the original version applied the Site URL PUT
**unconditionally**. `http://localhost:{config.platform.port}/metabase/` is only correct
for the plain local (no cloud, no custom domain) access pattern:

- **Cloud deployments** are fronted by Caddy with a real public domain/HTTPS, not
  `localhost:{port}` — writing the local value there would be actively wrong, not just
  unhelpful. Site URL also governs Metabase's own password-reset/invite email links and
  other absolute URLs it generates, not just JS asset paths, so this was a real regression
  risk for cloud specifically.
- Since `refresh_metabase_connection()` now applies this on every sync (not just once),
  an unconditional PUT would have meant **recurring corruption** for cloud projects, not a
  one-time no-op.

Fixed by adding `_should_apply_local_site_url(project_root, cloud_mode)`, checked *before*
any login/PUT is attempted in both call sites:
- `setup_metabase()` already receives `cloud_mode` as a parameter — short-circuits on it,
  no config/network lookup needed.
- `refresh_metabase_connection()` now calls `dango.config.helpers.is_cloud_mode()` first.
- Both also skip when the project is registered under a non-default domain via
  `platform/local/network.py`'s shared nginx routing (the local custom-domain feature).
  This is defense in depth rather than an active concern today: grep confirmed
  `NetworkConfig.register_project()` is currently only ever called from the `dango rename`
  CLI command, conditional on a routing.json entry nothing else creates — there's no live
  registration path for a *fresh* project. A project actually using this feature already
  has the same JS-asset-path bug via a different Site URL mismatch (nothing set Site URL
  at all before this fix), so this check isn't closing a new gap for such projects — it's
  making sure this PR doesn't start actively writing a wrong value where none was written
  before. Fully correct Site URL handling for local custom domains is still out of scope
  and documented as a known follow-up gap.

Also added while addressing this, since it changes how often the PUT runs:
- A `site_url_set` marker persisted to `.dango/metabase.yml`. Without it,
  `refresh_metabase_connection()` would repeat a login POST + PUT on every single sync
  forever (~20s of extra latency budget and retransmitting the admin password, for zero
  benefit once it's already correct). Written by whichever of `setup_metabase()` /
  `refresh_metabase_connection()` first successfully applies the fix; the marker makes it
  a one-time catch-up, not a recurring cost.
- `_set_metabase_site_url()` now returns `bool` (PUT succeeded or not) so callers can
  decide whether it's safe to persist the marker.
- `_safe_print()` guards this module's own status glyphs (✓/⚠) so a non-UTF-8 stdout can't
  turn an already-caught PUT failure into a new, uncaught `UnicodeEncodeError` from the
  fallback print itself (the original fallback branch's own `print()` call wasn't
  protected).
- Updated `_metabase_login()`'s docstring, which claimed "three" call sites — grep-verified
  the actual current list instead of leaving a stale count now that a fourth caller exists.

## Update 2: second independent review pass, three more fixes
A second review pass (same `code-review` skill, run fresh against the pushed fix above)
found three more real issues, now fixed, plus two items I'm flagging as known
limitations rather than fixing in this PR (see below):

- **The custom-domain check from Update 1 was dead code.** `_should_apply_local_site_url()`
  compared the registered domain to `f"{project_name}.dango"` to detect a "custom" domain
  — but `NetworkConfig.register_project()` (the only real caller, via `dango rename`)
  *always* registers under exactly that pattern for the *current* project name. That
  comparison could never be true, so the check silently never skipped anything for the one
  live registration path — defeating its own stated purpose. **Fixed:** any registration at
  all now disqualifies (a registered project is accessed via its routing.json `domain`
  entry, not `localhost:{port}`, regardless of what that domain string happens to be). New
  positive-control test (`test_skips_registered_project_even_with_default_domain_pattern`)
  proves the fixed behavior on the exact scenario that was broken — see
  `TestShouldApplyLocalSiteUrl` for direct unit tests on this function.
- **Recurring I/O contradicting the "one-time catch-up" framing.** The `site_url_set`
  marker check ran *after* `_should_apply_local_site_url()`'s heavier work (a full
  `project.yml` reparse, a `NetworkConfig()` construction touching `~/.dango/`), so an
  already-caught-up project still paid for that local I/O on every single sync forever —
  not the network I/O (no login/PUT happened), but still unnecessary recurring work
  contradicting the docstring's own claim. **Fixed:** extracted the whole catch-up flow
  into `_apply_metabase_site_url_catchup()`, which now checks the marker first (one cheap
  file read) before calling `_should_apply_local_site_url()` at all.
- Extracting that helper also resolves a code-quality finding from this same pass: the
  catch-up logic was inlined 6 levels deep (`for → try → try → if → if → if`) directly in
  `refresh_metabase_connection()`'s health-poll loop, unlike `_set_metabase_site_url()`
  which was already its own function — inconsistent extraction made the orchestration
  harder to follow. It's now a named, independently-testable function.
- Also addressed: `_safe_print()`'s docstring narrowed to accurately state it only guards
  this PR's own two new print calls, not the dozen-plus pre-existing unguarded ✓/⚠ prints
  elsewhere in this module (it was implicitly overclaiming module-wide protection);
  `dango/visualization/CLAUDE.md`'s stale "No dango module imports (isolated module)"
  claim updated to document the three lazy `dango.*` imports this PR adds and why;
  `metabase.py`'s line count corrected (1251 → 1775, now genuinely accurate) in the two
  root `CLAUDE.md` monolithic-files entries and `docs/file-exemptions.yml`, all stale
  after this PR's growth.

**Flagged as known limitations, not fixed in this PR** (would meaningfully expand scope
beyond this PR's bug):
- **Unsynchronized read-modify-write on `.dango/metabase.yml`.** `refresh_metabase_connection()`
  is reachable from four independent, non-serialized call sites (`dlt_runner.py`,
  `web/routes/dbt.py`, `platform/scheduling/jobs.py`, `cli/commands/transform.py`), and
  its new catch-up path does a read-modify-write of the whole credentials file with no
  lock. This is currently self-converging for `site_url_set` specifically (whichever
  writer runs, the field ends up `True`), but it's a lost-update pattern on a file that
  also holds live admin credentials, and `DbtLock` doesn't cover this file. Properly fixing
  this means adding file locking to a *pre-existing* unprotected-write pattern shared by
  `setup_metabase()`, `_link_metabase_admin()`, and now this catch-up path — a bigger,
  separate change than this PR's scope. Flagging for the coordinating chat to decide
  whether that's worth a dedicated follow-up task.
- **Blocking I/O in an async web handler.** `web/routes/dbt.py`'s `run_dbt_model_task` is
  `async def` and calls `refresh_metabase_connection()` synchronously with no thread-pool
  offload. That function already did significant blocking work before this PR (Docker
  `subprocess.run()` restart + up to 20s of health-polling) — a pre-existing architectural
  issue matching `CLAUDE.md`'s documented "single worker uvicorn" constraint. This PR's
  catch-up adds at most one more ~20s hit (one login POST + one PUT, both 10s timeout), but
  only *once* per project, on the first sync after upgrading, thanks to the `site_url_set`
  marker — not a new *recurring* cost. Not fixing the broader blocking-I/O-in-async-handler
  pattern here; noting it in case the coordinating chat wants it tracked separately.

## Update 3: third independent review pass — self-contained contracts, one latent bug fixed
A third review pass found smaller robustness/consistency issues plus one genuinely
unrelated pre-existing bug it happened to surface:

- `_apply_metabase_site_url_catchup()` claimed "never raises" but only held because its
  caller wrapped it — the function's own body wasn't self-contained. Fixed: the whole body
  is now in its own try/except, so the guarantee doesn't depend on caller behavior.
- `_should_apply_local_site_url()`'s fail-open `except: pass` silently swallowed every
  error, including ones worth knowing about (e.g. a corrupted `~/.dango/routing.json`).
  Now logs at debug level (fail-open behavior unchanged — still deliberate). Also switched
  from a full `ConfigLoader.load_config()` to the lighter `load_project_context()`, since
  only `project.name` is needed and the full load also parses/validates `sources.yml` for
  no benefit here.
- **Found and fixed a real, pre-existing, unrelated bug** while adding that debug log,
  directly in the function this PR touches: the final `except Exception as e:` block did
  `logger.warning("...", error=str(e), exc_info=True)` — `error=` isn't a valid kwarg for
  stdlib `logging.Logger.warning()` (confirmed live: this module's `logger` is
  `logging.getLogger`, not structlog). That line itself raised `TypeError`, silently
  replacing whatever the real exception was with an unrelated one, every time this path
  fired. Fixed to a valid call shape. New positive-control regression test
  (`test_unexpected_error_returns_gracefully_instead_of_masking_typeerror`) forces an
  exception and confirms the function now returns `(False, message)` instead of raising.
- Documented two accepted asymmetries in the `site_url_set` marker design directly in
  `_apply_metabase_site_url_catchup()`'s docstring: cloud/registered projects never get the
  "one-time" savings (intentional — lets them self-correct if their topology later becomes
  plain-local), and a project that becomes cloud/registered *after* already being marked
  has no automatic re-check (an extension of the already-disclosed local-custom-domain
  follow-up gap, not a new one — and, per the grep in Update 1/2, not currently reachable
  since the only live registration path only updates an existing entry, never creates one).
- `_metabase_login()`'s docstring corrected (it named `refresh_metabase_connection()` as a
  direct caller, but that call now goes through `_apply_metabase_site_url_catchup()`), and
  revised to acknowledge that this PR's own new credential-loading code is itself another
  copy of the "open metabase.yml, pull out admin.email/password" shape the docstring
  already flagged as worth consolidating — rather than silently exceeding its own stated
  threshold without saying so.

## Why this fix (not the nginx config)
Metabase computes its own JS chunk-loading URLs from its internal **Site URL** setting,
not from how a reverse proxy in front of it is configured. Dango proxies Metabase through
`dango/web/routes/metabase_proxy.py` (mounted at `/metabase`), whose `_build_response()`
does a pure byte-for-byte passthrough — no HTML/JS rewriting. Since nothing previously told
Metabase its Site URL is `/metabase/`-prefixed, it assumed basename `/` and generated
chunk-loading requests that 404/mis-route once the browser is actually at
`http://localhost:8800/metabase/...`.

`dango/templates/nginx.conf.j2` is confirmed dead code (not referenced anywhere in the
Python codebase) and is untouched by this PR.

A previous attempt (commit `a98072e`, reverted same day by `effb900`) set
`MB_SITE_URL: "http://localhost:3000"` as a docker-compose env var — missing the
`/metabase/` suffix and using Metabase's internal port (3000) instead of the
browser-visible web UI port (8800, `config.platform.port`). That attempt was reverted on
the incorrect conclusion that the warning is cosmetic; it is not (see symptom above), and
the value was wrong on two counts. This PR uses the full `/metabase/`-suffixed URL and the
correct web UI port, applied via the live Admin API rather than a docker-compose env var —
which also means it recomputes the URL correctly regardless of what port the project is
actually configured on, instead of hardcoding one, and now (post-review) gates on
deployment topology so it doesn't apply where it would be wrong.

**This PR is scoped to local-only.** It does not touch or verify cloud deployment's
Caddy-based proxying, per the task's explicit instruction not to assume the local fix
covers cloud — it now actively *avoids* writing anything for cloud deployments (see Update
above) rather than just not addressing them. Whether cloud needs its own equivalent fix
(presumably setting Site URL to the public domain, via a different code path) is an open
follow-up question, not something this PR answers either way.

## Verification
- `pytest -x`: 4670 passed, 30 skipped, 0 failed (full suite, no regressions)
- `python3 scripts/check_file_sizes.py --all`: 406 checked, 82 exempt, all within limits
- Unit tests in `tests/unit/test_metabase_site_url.py::TestSetupMetabaseSiteUrl` (via
  `setup_metabase()`):
  - `test_setup_metabase_sets_site_url` — PUT made to `{metabase_url}/api/setting/site-url`
    with `json={"value": "http://localhost:8800/metabase/"}` when
    `config.platform.port == 8800`; `.dango/metabase.yml`'s `site_url_set` ends up `True`
  - `test_setup_metabase_site_url_failure_does_not_block_setup` — non-200 PUT doesn't stop
    DuckDB connection / credential save, and `site_url_set` stays `False` (so the next
    sync retries it)
  - `test_setup_metabase_skips_site_url_in_cloud_mode` — `cloud_mode=True` skips the PUT
    entirely, no `NetworkConfig` lookup
  - `test_setup_metabase_skips_site_url_for_registered_project` — a registered project
    skips the PUT
- Direct unit tests on the gate itself,
  `tests/unit/test_metabase_site_url.py::TestShouldApplyLocalSiteUrl` (added in Update 2):
  - `test_skips_in_cloud_mode_without_any_io` — cloud_mode short-circuits before any
    ConfigLoader/NetworkConfig call
  - `test_skips_registered_project_even_with_default_domain_pattern` — **positive control**
    for the dead-check bug: a registration under exactly `f"{project_name}.dango"` (what
    `dango rename` always produces) is still correctly skipped
  - `test_skips_registered_project_with_a_different_domain` — also skips for an unrelated
    domain string
  - `test_applies_for_unregistered_project` — the common case, no registration → `True`
  - `test_fails_open_when_config_cannot_be_loaded` — `ConfigNotFoundError` → fails open
- Unit tests in `tests/unit/test_platform_startup.py::TestRefreshMetabaseConnection`:
  - `test_unexpected_error_returns_gracefully_instead_of_masking_typeerror` — **positive
    control** for the pre-existing latent logger bug found and fixed in Update 3: forces
    an exception inside the try block, confirms `(False, message)` instead of a masking
    `TypeError`
  - `test_reapplies_site_url_on_successful_restart` — logs in with saved admin
    credentials, PUTs the correct site-url, and persists `site_url_set: true`
  - `test_site_url_failure_does_not_block_refresh` — login/PUT failure still returns
    `(True, None)`, and `site_url_set` is not written (so it retries next sync)
  - `test_skips_site_url_when_already_marked_set` — no login, no PUT, when the marker is
    already `True`
  - `test_skips_site_url_in_cloud_mode` — `is_cloud_mode()` True skips login+PUT entirely
  - Both pre-existing container-name-resolution tests (`test_uses_hash_based_container_name`,
    `test_returns_false_when_container_not_running`) still pass unmodified
- **Manual browser verification (requires Docker) was not performed in this session.**
  Ports 8800/3000 were already in use by another active local Metabase instance at the time
  of this session (likely beta-1 or a concurrent session) — per project rules, sessions must
  not run live Docker/browser tests against shared/already-running instances; that
  verification is the coordinating chat's job before merge (sequential, one PR at a time).
  Please verify in a real browser before merging: fresh `dango init` + `dango start`, open
  `http://localhost:<configured-port>/metabase/`, confirm no `Uncaught SyntaxError` in the
  console and the dashboard renders (not blank). Also worth spot-checking the
  `refresh_metabase_connection()` path specifically: run a sync on a project whose
  `.dango/metabase.yml` was created *before* this branch (simulating an already-configured
  project), confirm the Site URL gets corrected after the post-sync restart, and confirm a
  *second* sync does not repeat the login/PUT (marker skip).

## Regression check
- Site-url-set failure confirmed non-blocking on both call sites (see tests above).
- Grepped all callers: `setup_metabase()` is only called from
  `platform/common/startup.py` and `cli/init.py`; `refresh_metabase_connection()` is called
  from `dlt_runner.py`, `web/routes/dbt.py`, `platform/scheduling/jobs.py`, and
  `cli/commands/transform.py` (post-sync/post-dbt paths). None of these depend on internals
  this change touches.
- **Cloud deployments:** now actively excluded (see Update above), not just undocumented —
  this closes what an independent review flagged as a genuine regression risk.
- **Local custom-domain projects:** known, documented follow-up gap — see Update above for
  why this is pre-existing (not introduced by this PR) and why the defense-in-depth check
  is enough for now rather than a full fix.
- `dango/templates/nginx.conf.j2` and `platform/local/network.py` are untouched (only read
  from, via `NetworkConfig.get_project_info()`), as specified.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -208,7 +208,7 @@ dango/                          # Python package source
 │   └── static/                 # Frontend HTML/CSS/JS
 │
 ├── visualization/              # Level 2 — Metabase integration
-│   ├── metabase.py             # Metabase API (1251 lines)
+│   ├── metabase.py             # Metabase API (1775 lines)
 │   └── dashboard_manager.py    # Dashboard export/import (1117 lines)
 │
 ├── platform/                   # Level 2 — Docker, network, file watcher, scheduling
@@ -343,7 +343,7 @@ Full exemption registry: [`docs/file-exemptions.yml`](docs/file-exemptions.yml)
 | `ingestion/dlt_runner.py` | 2885 | — (exempt, too risky) |
 | `ingestion/sources/registry.py` | 2340 | — (metadata-only) |
 | `cli/source_wizard.py` | 2610 | — |
-| `visualization/metabase.py` | 1251 | — |
+| `visualization/metabase.py` | 1775 | — |
 | `cli/init.py` | 1585 | — |
 | `visualization/dashboard_manager.py` | 1117 | — |
 | `cli/commands/platform.py` | 1136 | — (extracted from main.py by TASK-005) |
--- a/dango/visualization/CLAUDE.md
+++ b/dango/visualization/CLAUDE.md
@@ -24,7 +24,14 @@ Integrates with Metabase for dashboard provisioning, auto-setup, schema synchron
 ## Dependencies
 
 **Imports from:**
-- No dango module imports (isolated module). Uses `requests` to call Metabase API, reads credentials from `.dango/metabase.yml` at runtime.
+- Mostly isolated (uses `requests` to call Metabase API, reads credentials from
+  `.dango/metabase.yml` at runtime), with one exception: `metabase.py`'s Site URL
+  handling (1.0.8-W — `_should_apply_local_site_url()`, `_apply_metabase_site_url_catchup()`)
+  lazy-imports `dango.config.ConfigLoader`, `dango.config.helpers.is_cloud_mode`, and
+  `dango.platform.local.network.NetworkConfig` to check deployment topology (cloud mode,
+  local shared-nginx registration) before writing a `localhost:{port}`-based Metabase Site
+  URL — all Level 0–2 imports, consistent with this module's Level 2 position in the
+  dependency hierarchy.
 - `DASHBOARD_QUERIES`' SQL (not Python imports) reads the `_dango_meta` schema tables written by `dango.utils.pipeline_health.materialize_pipeline_health()` (1.0.8-DASH-1) — `cli/commands/dashboard.py` calls that function (and `refresh_metabase_connection()`, below) before provisioning, not this module.
 
 **Used by:**
--- a/dango/visualization/metabase.py
+++ b/dango/visualization/metabase.py
@@ -265,6 +265,42 @@ def _metabase_login(
     return response.json().get("id")
 
 
+def _set_metabase_site_url(
+    session: requests.Session,
+    metabase_url: str,
+    headers: dict[str, str],
+    project_root: Path,
+) -> None:
+    """PUT /api/setting/site-url so the local /metabase/ proxy serves correctly-pathed
+    JS assets — Metabase computes chunk-loading URLs from this setting, and the local
+    proxy (metabase_proxy.py) does not rewrite response bodies to compensate.
+
+    Shared by setup_metabase() (once, at initial setup) and refresh_metabase_connection()
+    (every sync-triggered restart — the only path that ever reaches an already-configured
+    project, i.e. one set up before this fix shipped). The PUT is idempotent — setting the
+    same value repeatedly is harmless — so callers do not need to check "was it already
+    set" before calling this. Never raises; failure is printed as a warning only.
+    """
+    try:
+        from dango.config import ConfigLoader
+
+        config = ConfigLoader(project_root).load_config()
+        web_port = config.platform.port
+        site_url = f"http://localhost:{web_port}/metabase/"
+        site_url_resp = session.put(
+            f"{metabase_url}/api/setting/site-url",
+            headers=headers,
+            json={"value": site_url},
+            timeout=10,
+        )
+        if site_url_resp.status_code == 200:
+            print(f"  ✓ Metabase Site URL set to {site_url}")
+        else:
+            print(f"  ⚠ Could not set Metabase Site URL: {site_url_resp.status_code}")
+    except Exception as e:  # noqa: BLE001
+        print(f"  ⚠ Could not set Metabase Site URL: {e}")
+
+
 class MetabaseProvisioner:
     """
     Provisions Metabase dashboards via API
@@ -971,6 +1007,9 @@ def setup_metabase(
 
         # At this point, we have headers with session token from either path
 
+        # Set Site URL so the local /metabase/ proxy serves correctly-pathed JS assets.
+        _set_metabase_site_url(session, metabase_url, headers, project_root)
+
         # Check for existing DuckDB connection to prevent duplicates
         existing_db_id = None
         db_name = f"{org_name} Analytics"
@@ -1561,6 +1600,32 @@ def refresh_metabase_connection(
             try:
                 response = session.get(f"{metabase_url}/api/health", timeout=1)
                 if response.status_code == 200:
+                    # Re-apply Site URL on every restart, not just at initial setup.
+                    # This is the only path that ever reaches a project whose
+                    # .dango/metabase.yml was created before this fix shipped —
+                    # setup_metabase() only runs once and won't retroactively fix
+                    # those projects. Best-effort: login + PUT are wrapped so a
+                    # failure here never turns a successful restart into a failure.
+                    try:
+                        creds_file = project_root / ".dango" / "metabase.yml"
+                        if creds_file.exists():
+                            with open(creds_file) as f:
+                                creds = yaml.safe_load(f) or {}
+                            admin = creds.get("admin", {})
+                            email = admin.get("email")
+                            password = admin.get("password")
+                            if email and password:
+                                session_id = _metabase_login(session, metabase_url, email, password)
+                                if session_id:
+                                    _set_metabase_site_url(
+                                        session,
+                                        metabase_url,
+                                        {"X-Metabase-Session": session_id},
+                                        project_root,
+                                    )
+                    except Exception:  # noqa: BLE001
+                        pass  # Not critical — site URL refresh is best-effort
+
                     return (True, None)
             except requests.exceptions.RequestException:  # noqa: BLE001
                 pass
--- a/dango/visualization/metabase.py
+++ b/dango/visualization/metabase.py
@@ -284,11 +284,20 @@ def _should_apply_local_site_url(project_root: Path, cloud_mode: bool) -> bool:
 
     - **Cloud deployments** — Caddy fronts Metabase there with the real public
       domain/HTTPS, not ``localhost:{port}``.
-    - **Local custom-domain projects** — ``platform/local/network.py``'s shared
-      nginx routing can register a project under ``{project_name}.dango``
-      (or another domain) instead of ``localhost:{port}``.
+    - **Local shared-nginx-registered projects** — ``platform/local/network.py``'s
+      shared nginx routing serves a registered project at its routing.json
+      ``domain`` entry (``{project_name}.dango`` by default, or a custom domain),
+      not ``localhost:{port}`` — *any* registration means this module's computed
+      URL is wrong for that project, not just a non-default one. (An earlier
+      version of this check compared the registered domain to
+      ``f"{project_name}.dango"`` to detect a "custom" domain — but
+      ``NetworkConfig.register_project()`` always registers under exactly that
+      pattern for the current project name, so that comparison could never be
+      true for the only real caller and silently never skipped anything. Fixed
+      during review to treat any registration as disqualifying, regardless of
+      what the domain string is.)
 
-    The custom-domain check is defense in depth rather than an active concern
+    This registration check is defense in depth rather than an active concern
     today: grepping the codebase confirmed ``NetworkConfig.register_project()``
     is currently only ever called from the ``dango rename`` CLI command,
     conditional on a routing.json entry that nothing else creates — there is no
@@ -297,7 +306,9 @@ def _should_apply_local_site_url(project_root: Path, cloud_mode: bool) -> bool:
     mismatch (nothing set Site URL at all before this fix), so this check isn't
     closing a new gap for such projects — it's just making sure this PR doesn't
     start actively writing a wrong value where none was written before. Fully
-    correct Site URL handling for local custom domains is out of scope here.
+    correct Site URL handling for locally-routed/custom-domain projects (e.g.
+    writing ``http://{domain}/metabase/`` instead of skipping) is out of scope
+    here.
 
     Fails open (returns True) if project name / routing lookup can't be
     completed, matching the common case (no registration exists) rather than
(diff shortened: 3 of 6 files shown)


aaronteoh:
## Manual verification (1.0.8-T5 fresh-install pass) — found and fixed one more bug

Set up a scratch project (`tests/scratch-w-verify/`, isolated ports) running this branch's actual code via `PYTHONPATH=<worktree>` to do the manual browser check this PR was blocked on (confirm no blank page / console errors at the actual proxied Metabase URL).

**Found:** `dango start` reported `⚠ Could not set Metabase Site URL: 204` on every restart. Root cause: `_set_metabase_site_url()` checked `site_url_resp.status_code == 200`, but Metabase's `/api/setting/*` PUT endpoints return **204 No Content** on success, not 200 — confirmed live. The sibling `anon-tracking-enabled` PUT in the same file already handles this correctly via `raise_for_status()`; this one used a strict equality check instead, which silently no-op'd the entire fix on every real invocation.

**Fixed:** switched to `site_url_resp.ok` (accepts any 2xx). Pushed as `8b3b001`.

**Live verification after the fix:**
- CLI output: `✓ Metabase Site URL set to http://localhost:8900/metabase/`
- Direct API check (`GET /api/setting/site-url` with a real session): confirmed the value persisted in Metabase itself, not just the CLI's own success message
- Browser: navigated to the proxied `/metabase` URL — page renders fully styled (not blank), zero console errors
- Network tab: all 15 asset requests (`app/dist/*.js`, `*.css`, fonts, manifest) resolve **200** under the correct `/metabase/`-prefixed path — this is the literal mechanism of the original blank-page bug, now confirmed fixed
- Added a regression test (`test_setup_metabase_site_url_204_is_treated_as_success`) and fixed two existing test mocks whose unconfigured `MagicMock.ok` would have silently masked this exact bug (auto-generated Mock attributes are truthy regardless of the simulated status code)
- Full suite: `pytest tests/unit/test_metabase_site_url.py tests/unit/test_platform_startup.py` — 57 passed

This was the last outstanding blocker on this PR. Merging once CI is green on the new commit.

PR #496, opened 2026-09-13
1.0.8-REL-1: Release-only clean-flow CI job

aaronteoh:
## Summary
- Added `.github/workflows/release-readiness.yml` — manually-triggered
  (workflow_dispatch only) job for release-only clean-flow checks
- Added `tests/integration/test_release_readiness.py` — real-nginx-proxy
  Metabase asset check, and second-source schema-sync race check

## Deviation from the task doc (re-verified against live code, as instructed)
The task doc's "what to read first" pointed at `dango/templates/nginx.conf.j2`
as "the real nginx proxy" a local `dango start` project routes `/metabase/*`
through. Grepping the codebase (and confirmed by `dango/templates/CLAUDE.md`
itself, which calls that file "not yet consumed (placeholder)") shows this file
is dead code — nothing renders or loads it. The actual local `/metabase/*`
proxy is FastAPI's own `dango/web/routes/metabase_proxy.py`, mounted directly
on `dango.web.app:app` — the exact code path 1.0.8-W's Site URL fix targets.
Both tests exercise that real app (started via
`dango.cli.helpers.process_manager.start_fastapi_server()`, the same real
uvicorn-subprocess entry point `dango start` itself calls) rather than nginx.

## UPDATE 2026-09-16: the underlying race is now fixed, test passes for real

This PR originally surfaced a real restart-readiness race in
`refresh_metabase_connection()` (see "Original finding" below) rather than a
bug in this PR's own code. That race took 3 more attempts to actually close
(1.0.8-Q11 → Q14, reverted as a live regression → Q16, caught by its own
mandatory live-verification gate before merge → **Q17, merged**, PR #505 —
full incident history in `BUGS-FOUND.md`). A second, separate bug
(`sync_metabase_schema()`/`refresh_metabase_connection()` always defaulting to
`localhost:3000` regardless of a project's configured `metabase_port`) was
found and fixed along the way (PR #506) — this test's own scratch project uses
a non-default port, so it depended on that fix too.

This branch has been rebased onto current `v1.0.8` (includes both fixes) and
this test's own separate bug — `_sync_until_table_visible()` checked
`source_name in names` directly, which can never match since dbt staging
tables are always named `stg_{source.name}__{table_name}` — is now fixed too
(switched to a `startswith("stg_{source_name}__")` check).

**Re-run for real against the final merged fix:**
```
DANGO_RUN_DOCKER_TESTS=1 pytest tests/integration/test_release_readiness.py -v
[...]

PR #502, opened 2026-09-14
1.0.8-Q14: Widen Metabase login-readiness non-200 retry budget

aaronteoh:
## Summary
- Widen `_wait_for_metabase_login_ready()`'s retry budgets to ~60s each (`max_attempts` 10 → 60, `max_non_200_attempts` 5 → 60 — both, since the two share one loop and only raising one is a no-op past the smaller value), per real diagnostic evidence from REL-1 (PR #496) that a restart immediately following a real sync pipeline needs more than 5-10s to boot
- Update the function's docstring to explain the new budget and the accepted tradeoff (a genuinely wrong password now also takes longer to report failure at this call site — `dango auth metabase-status`/`metabase-repair` from 1.0.8-Q13 cover that diagnosis path instead)
- Update 3 existing test call-count assertions to match the new default (no new tests — existing coverage already exercises this path)

## Scope correction during implementation
The original task spec called for raising only `max_non_200_attempts` (5 → 60) while leaving `max_attempts` at 10. Before editing, I read the actual loop body and found both retry types share a single `for _ in range(max_attempts):` loop — `max_non_200_attempts` can only end the loop *early* (when smaller than `max_attempts`), never extend it past `max_attempts` iterations. Raising only `max_non_200_attempts` would have been a no-op beyond `max_attempts` itself (proved via a test run: expected `call_count == 61`, got `call_count == 11`). Confirmed with the task owner and corrected scope to raise both defaults to 60, without restructuring the loop.

## Verification
- `pytest -x`: 4702 pass, 0 fail, 30 skipped (full suite)
- `pytest tests/unit/test_platform_startup.py`: 48 pass, 0 fail
- `ruff check` / `ruff format --check` / `mypy`: clean

## Regression check
- `refresh_metabase_connection()`'s outer `/api/health` loop (`max_attempts = 20`, a different function) confirmed unchanged
- `grep -n "call_count == [0-9]"` reviewed — only the 3 intended tests changed (`call_count == 3` at two other unrelated locations left untouched)

https://claude.ai/code/session_01Nn6RWFt5T69QR6gvQ27igT


aaronteoh:
## Independent code review (self-requested, code-review skill, level=high)

Ran an independent review pass against this PR. Six findings came back; evaluated each individually rather than accepting/dismissing wholesale.

### Fixed directly (cheap, unambiguous, no design decision needed)
1. **`refresh_metabase_connection()`'s outer `# Wait for Metabase to come back up (max 20 seconds)` comment** was already slightly stale pre-PR and is now materially misleading: once `/api/health` returns 200, the function calls `_wait_for_metabase_login_ready()` once, which can now add up to ~60s more before returning — not captured by the "20 seconds" framing at all. Fixed in the comment (metabase.py, the outer loop).
2. **`test_site_url_failure_does_not_block_refresh()`'s docstring** still claimed the retry path "must complete quickly (a bounded few seconds, not the full ~10s connection-error budget)" — directly contradicted by this same PR's own `call_count` assertion change (6→61) five lines below. Updated the docstring to describe the 1.0.8-Q14 widening instead of contradicting the test's own updated assertion.

Both pushed as a follow-up commit; `ruff`/`mypy`/full `pytest -x` still green after.

### Flagging for discussion, not unilaterally acting on
3. **Most significant finding:** `refresh_metabase_connection()` calls `_wait_for_metabase_login_ready()` synchronously from inside `run_dbt_model_task()` (`dango/web/routes/dbt.py:216`), which is an `async def` function FastAPI's `BackgroundTasks` awaits directly on the event loop (not a thread pool) — confirmed by reading `dbt.py:44-96` and the `BackgroundTasks.add_task` call site. On Dango's single-worker uvicorn (a deliberate architecture choice per project notes), this blocks *all* other HTTP/WebSocket traffic server-wide for however long the readiness poll runs. This blocking call already existed before this PR; this PR increases its ceiling from ~10-15s to up to ~60s in the common case. This is a real, pre-existing architectural issue that this PR's change makes meaningfully worse for the exact scenario (restart-after-sync) it targets. Fixing it properly (e.g. `asyncio.to_thread`/`run_in_executor` around the call, and auditing other callers of `refresh_metabase_connection()` for the same pattern) is a bigger change than this PR's stated scope ("only change two defaults, don't restructure"). Raising it here rather than fixing it unilaterally, since it's a decision about scope, not a mechanical correction.
4. **Docstring worst-case math**: the new docstring text (dictated verbatim in the task spec) says the non-200/connection budgets are "~60s by default." Each attempt calls `_metabase_login()` with its own default `timeout=10`, uncapped by the outer 1s sleep — so a slow-but-not-instantly-refused connection could in theory push the real worst case toward `max_attempts * (10+1)s ≈ 660s`, not ~60s. This is a pre-existing approximation pattern (the old docstring had the same proportional gap at smaller absolute numbers, ~10s claimed vs ~110s theoretical worst case) rather than a new bug this PR introduces — flagging as a minor accuracy nit on text that was specified verbatim by the task, not rewriting unilaterally.
5. Reviewer also flagged the lack of an enforced relationship between `max_attempts` and `max_non_200_attempts` (no clamp/assertion) as a future-footgun. The corrected task spec explicitly considered and accepted this exact tradeoff ("leave it in place... it's still meaningful for any future caller that passes a smaller `max_non_200_attempts` explicitly... removing it is out of this task's scope") — treating this as already-decided, no action taken.
6. Reviewer also flagged the docstring's length/duplication with the PR description as a maintainability nitpick — doesn't clear the bug-that-bites/correctness bar, not acting on it.

Not merging — per project workflow, the coordinating chat verifies and merges.

PR #503, opened 2026-09-14
1.0.8-Q15: Fix event-loop-blocking Metabase refresh in dbt web route

aaronteoh:
## Summary
- Wrap `refresh_metabase_connection()`'s call in `run_dbt_model_task()` (`web/routes/dbt.py`) with `asyncio.to_thread()` so it no longer blocks the single-worker FastAPI event loop for up to ~60s (widened by 1.0.8-Q14) while running
- Confirmed the other 2 callers of `refresh_metabase_connection()` relevant to this async-event-loop concern are not affected by this same issue (subprocess-isolated / already thread-pool-executed) — not touched: `dlt_runner.py`'s `run_sync()`, `platform/scheduling/jobs.py`'s `_run_coalesced_dbt()`

## Verification
- `pytest -x`: 4703 passed, 30 skipped, 0 failed (full suite, foreground)
- `ruff format --check dango/`: all files already formatted
- New regression test `test_metabase_refresh_does_not_block_event_loop`: manually confirmed it fails with the fix reverted, passes with the fix applied

### Note on the regression test's enforcement mechanism
The task spec's originally-proposed version of this test wrapped the concurrency check in `asyncio.wait_for(timeout=3)`, expecting a `TimeoutError` if the refresh call blocked the loop directly. When manually verifying the revert-and-confirm-failure step, that version did **not** fail when reverted — it just ran slower and still passed. Root cause: `asyncio.wait_for`'s own timeout callback can't fire while the event loop is synchronously blocked inside the un-fixed code path — the loop is occupied by the very call the timeout is supposed to catch, so `wait_for` can never preempt it. The test was adjusted to assert wall-clock elapsed time (`< 1.0s`) around the `asyncio.gather()` call instead, which correctly discriminates both cases without depending on a timeout race against the loop-blocking bug itself. The blocking mock's own internal `threading.Event.wait(timeout=2)` still bounds worst-case runtime so the test can never hang.

## Regression check
- The two other `refresh_metabase_connection()` call sites relevant to this async-blocking concern (`dlt_runner.py`, `platform/scheduling/jobs.py`) untouched, confirmed via grep


aaronteoh:
## Independent review (self-authored PR — findings only, not an approval; coordinating chat merges)

Reviewed the full diff plus surrounding context (not just the changed lines):

**Correctness**
- The `await asyncio.to_thread(...)` call sits inside the same `try/except Exception as e` / `finally: lock.release()` block that already wrapped the original synchronous call — exception propagation and lock release are unchanged. Verified by reading the full `run_dbt_model_task()` body, not just the diff.
- `refresh_metabase_connection()` (`dango/visualization/metabase.py:1933`) uses only `subprocess.run`, a locally-scoped `requests.Session()`, and stdlib logging — no asyncio-only calls, no shared mutable global state. Safe to run off the event loop thread; contextvars (structlog context) propagate into the worker thread automatically per `asyncio.to_thread`'s documented behavior.

**Convention consistency**
- `asyncio.to_thread` is already the established idiom for blocking calls throughout this codebase (80+ call sites: `web/routes/catalog.py`, `notebooks.py`, `governance.py`, `health.py`, `auth/metabase_bridge.py`, `cli/commands/mcp_server.py`, etc.). This PR applies an existing, well-trodden pattern to a call site that had been missed — not introducing anything novel.

**Minor, non-blocking note**
- The new test's `assert elapsed < 1.0` wall-clock threshold has a comfortable margin against the bounded 2s failure floor, and the passing case should complete in tens of milliseconds — but wall-clock assertions in CI always carry some inherent flakiness risk under runner contention. Not severe enough to block merge; flagging so it's not a surprise if it ever flakes on a loaded runner.

**Out of scope, not raised as a finding**
- `project_root = get_project_root()` is redundantly reassigned right before the refresh call (pre-existing, untouched by this diff).

No blocking issues found.

PR #508, opened 2026-09-17
1.0.8-AC: Cover remaining Metabase staleness callers

aaronteoh:
## Summary
- 4 remaining sync_metabase_schema() callers (metabase_cmd.py, model.py, source.py,
  initial_sync.py) now call refresh_metabase_connection() first and thread
  existing_session_id through, matching the pattern already used by the 3 previously
  fixed callers (dlt_runner.py, jobs.py, transform.py)
- Corrected a misleading comment in source.py claiming Metabase is always restarted
  before this call site — true on cloud, a documented no-op on local
- initial_sync.py's _refresh_metabase() is now wrapped in asyncio.to_thread — it was
  previously blocking the single-worker event loop for the full container-restart
  duration during cloud initial-sync onboarding

## Verification
- `pytest -x`: 4725 passed, 32 skipped, 0 failed (full suite)
- `ruff format --check dango/`: passed
- `mypy dango/`: Success: no issues found in 241 source files
- `grep -rn "sync_metabase_schema(" dango/ --include="*.py"` (excluding the definition
  file `visualization/metabase.py`), showing all 7 real callers passing
  `existing_session_id=`:
  ```
  dango/ingestion/dlt_runner.py:3195:                    if sync_metabase_schema(project_root, existing_session_id=mb_session_id):
  dango/web/routes/initial_sync.py:313:            sync_metabase_schema(project_root, existing_session_id=mb_session_id)
  dango/platform/scheduling/jobs.py:392:                sync_metabase_schema(project_root, existing_session_id=mb_session_id)
  dango/cli/commands/metabase_cmd.py:371:        success = sync_metabase_schema(
  dango/cli/commands/metabase_cmd.py:372:            project_root, metabase_url, existing_session_id=mb_session_id
  dango/cli/commands/model.py:329:                if mb_ok and sync_metabase_schema(project_root, existing_session_id=mb_session_id):
  dango/cli/commands/transform.py:133:            if sync_metabase_schema(project_root, existing_session_id=mb_session_id):
  dango/cli/commands/source.py:1131:            if mb_ok and sync_metabase_schema(project_root, existing_session_id=mb_session_id):
  ```
- New unit tests added (all passing): `test_cli_commands.py::TestMetabaseRefreshCommand`,
  `TestModelRemoveMetabaseRefresh`; `test_cli_sync.py::test_source_sync_metabase_refresh_calls_refresh_connection_first`;
  new file `test_initial_sync_metabase_refresh.py` (split out of `test_initial_sync.py`,
  which was already at the repo's 500-line hard file-size limit — see note below)

## Regression check
- dango metabase refresh now restarts the container before syncing (latency increase,
  intentional trade-off — see task Background: correctness over speed for an explicit,
  infrequent command)
- source.py's call remains in place as a fallback for the dbt-failed-but-sync-succeeded
  case, not removed as redundant

## Note on test file placement
`tests/unit/test_initial_sync.py` was at 490/500 lines (the repo's enforced hard cap,
`scripts/check_file_sizes.py`) before this PR. Adding the new Metabase-refresh tests
in place would have pushed it to ~579 lines and failed the pre-commit `file-size-check`
hook. Since those tests only need `_refresh_metabase()`/`_run_initial_sync()` — not
`test_initial_sync.py`'s FastAPI app/client fixtures — they were split into a new file,
`tests/unit/test_initial_sync_metabase_refresh.py`, instead of trimming existing
coverage to make room.


aaronteoh:
## Independent review (code-review skill, effort: high)

Ran an independent review pass against this branch. Findings, and disposition against the task spec (`v1.0.x-planning/1.0.8/AC-metabase-callers-staleness.md`):

**1. `source.py` (line 1130) and `initial_sync.py` (line 311): both now trigger a redundant second Metabase container restart on the common path.**

Verified independently:
- `run_sync()` (`dlt_runner.py:2888`, `skip_dbt` param docstring: *"If True, skip dbt run/docs/Metabase refresh (caller handles separately)"*) already restarts Metabase + syncs schema internally (lines 3182-3196) whenever dbt succeeds.
- `source.py`'s `sync()` calls `run_sync()` without `skip_dbt=True`, then this PR adds a second `refresh_metabase_connection()` + `sync_metabase_schema()` call right after — a second restart on the happy path.
- `initial_sync.py`'s `_sync_single_source()` also calls `run_sync()` without `skip_dbt=True`, so the same internal refresh fires after every source during the per-source loop. This PR's `_refresh_metabase()` change upgrades the post-loop call from a cheap `sync_metabase_schema()`-only call to a full `refresh_metabase_connection()` restart, adding one more restart at the end.

Both are real, but they're the exact trade-off the task's own Background section pre-cleared: *"Redundant restarts are an already-accepted cost in this codebase, not a new trade-off to resolve here. ... This task's fix to the **other 4 callers** follows the same accepted pattern — do not attempt to deduplicate restarts across call sites as part of this task."* That sentence covers all 4 call sites this PR touches, not just `source.py`. No code change made — this was a decision already made at task-authoring time, not a new discovery, and I'm not overriding it unilaterally.

**2. Shared `refresh_and_sync_metabase()` helper to remove the 7-call-site copy-paste.**

Good simplification idea in principle, but the same sentence above explicitly rules it out for this task: *"that's a separate, larger optimization, not a correctness fix."* Not actioned here — worth a follow-up task if the duplication becomes a maintenance problem (it's already what let 4 callers drift out of sync with the pattern once).

**Conclusion:** no changes made in response to this review — both findings are accurate but already-accepted, already-documented trade-offs per the task spec, not new bugs. Flagging here per review discipline rather than silently dropping them.

PR #509, opened 2026-09-17
1.0.8-AB: Fix prompt redraw duplication on paste

aaronteoh:
## Summary
- `dango source add`'s "Spreadsheet ID or URL" prompt now uses click.prompt() instead
  of inquirer.Text, avoiding a confirmed upstream inquirer 3.4.1 renderer bug where
  pasting a value longer than the terminal width causes the redraw loop to duplicate
  the line dozens of times (root cause read directly from inquirer's own renderer
  source `render/console/__init__.py` — not a version-bump fix, 3.4.1 is the latest
  release and still has this behavior)
- Scoped to this one field only (the only confirmed reproduction)
- New tests live in their own file, `tests/unit/test_source_wizard_spreadsheet_prompt.py`
  — see "Test file placement" below
- Ctrl+C at this prompt is now caught and handled consistently with every other field
  — see "Ctrl+C fix" below, added after independent review caught a real regression

## Verification
- `pytest -x`: 4719 passed, 32 skipped, 0 failed
- `ruff format --check dango/`: all files already formatted
- Manual (paste behavior): reproduced the actual bug and the fix side-by-side under a
  real pty (80-column terminal), driving the exact code paths as real subprocesses
  (not mocked):
  - **Before (inquirer.Text, the code every other string param still uses):** pasted
    `https://docs.google.com/spreadsheets/d/1a2b3c4d5e6f7g8h9i0jklmnopqrstuvwxyz/edit#gid=0`
    as a single `write()` call. Captured raw terminal output shows the
    `[?] Spreadsheet ID or URL: ...` line re-rendered **87 times**, each redraw left on
    screen as a new line — the exact "dozens of duplicate/repeated redraw lines"
    behavior from the bug report.
  - **After (this fix, via `click.prompt()`):** same paste, same 80-column pty. Output
    shows a single clean `Spreadsheet ID or URL: <url>` line, prompt text appears
    exactly once, correct value returned.
- Manual (Ctrl+C behavior): drove the **actual `dango source add` CLI end-to-end**
  (not a unit test) through a real pty with proper session/controlling-terminal setup
  (`setsid()` + `TIOCSCTTY`, so the kernel's Ctrl+C→SIGINT delivery works exactly like
  a real terminal) — selected Google Sheets, named the source, reached the
  "Spreadsheet ID or URL" prompt, and sent a real Ctrl+C keystroke:
  - **Before the Ctrl+C fix:** `Spreadsheet ID or URL: ^C` immediately followed by a
    bare `❌ Error: ` (empty message — `str(click.Abort())==''`), then the wizard's
    normal `Source not added` / `Aborted!` cancellation text. The spurious error line
    made a clean cancel look like a crash.
  - **After the Ctrl+C fix:** `Spreadsheet ID or URL: ^C` goes straight to
    `Source not added` / `Aborted!` — no error line, identical to what every other
    field in the wizard produces on Ctrl+C.

## Regression check
- Downstream code receiving spreadsheet_url_or_id (`_fetch_google_sheets`,
  `_save_source`) still receives a plain string, unchanged — verified via
  `grep -rn "spreadsheet_url_or_id" dango/cli/source_wizard.py` (3 call sites,
  all still treat the value as a plain string) and via the pty reproduction above.
- Ctrl+C during this prompt now returns `None` from `_prompt_parameter()`, exactly
  like every other field's `if not answers: return None`, so it flows through
  `_collect_parameters()`'s `if value is None: return None` → `run()`'s
  `if params is None: return False` → the CLI's existing `Source not added` /
  `click.Abort()` → `Aborted!` sequence, with no exception ever reaching `run()`'s own
  `except KeyboardInterrupt:`/`except Exception as e:` clauses. Covered by
  `test_prompt_parameter_spreadsheet_url_ctrl_c_returns_none` (flipped from an earlier
  version of this test that asserted the buggy behavior — Abort propagating — as if it
  were correct; an independent review caught this).

## Test file placement
The tests were initially added to `tests/unit/test_source_wizard.py`, but that pushed
it from 424 to 524 lines, past the 500-line pre-commit gate. First pass added an
exemption to `docs/file-exemptions.yml` — reviewed and rejected: that registry already
carries ~15 entries and an exemption should be a logged, deliberate trade-off for
genuinely inseparable code, not the default move when a focused split is available (a
different session hit the identical problem on a different file in this same batch and
split rather than exempted). Reverted the exemption entirely and moved the tests into
`tests/unit/test_source_wizard_spreadsheet_prompt.py`.

## Ctrl+C fix
An independent code review (re-scoped after an earlier run mistakenly reviewed an
unrelated worktree — corrected before anything was acted on) found that `click.prompt()`
converts Ctrl+C into `click.Abort()` (a `RuntimeError`/`Exception`, not
`KeyboardInterrupt`), which the original version of this branch left unguarded. That
`Abort` skipped `run()`'s `except KeyboardInterrupt:` clause and landed in the generic
`except Exception as e:`, printing a bare `❌ Error: ` before the wizard's normal
cancellation message. Fixed by catching `click.Abort` locally and returning `None`,
matching the `if not answers: return None` convention every other field already uses.
Confirmed live (see Verification above) and via a flipped unit test.


aaronteoh:
## Independent review

Verified both findings against the PR head commit and the surrounding codebase (not just read — traced the actual control flow and checked click's own source).

### 1. Ctrl+C on this prompt produces a confusing bare error instead of a clean cancel (real regression)

`click.prompt()` catches `KeyboardInterrupt`/`EOFError` internally and raises `click.Abort()` (confirmed via `inspect.getsource(click.termui.prompt)`). `click.exceptions.Abort` subclasses `RuntimeError`/`Exception`, **not** `KeyboardInterrupt`, and `str(Abort())` is `''` (confirmed by instantiating it).

`run()`'s exception handling is:
```python
except KeyboardInterrupt:
    console.print("\n\n[yellow]Wizard cancelled[/yellow]")
    return False
except Exception as e:
    console.print(f"\n[red]❌ Error: {escape(str(e))}[/red]")
    return False
```
Every other field cancels via `inquirer.prompt()` returning `None` on Ctrl+C, which flows cleanly through `_collect_parameters`'s `if value is None: return None` → `run()`'s `if params is None: return False` — no exception at all. This new field instead raises `click.Abort`, which isn't caught anywhere in `_prompt_parameter`/`_collect_parameters`, skips the `except KeyboardInterrupt` clause, and lands in `except Exception as e`, printing a bare `❌ Error: ` (empty message, since `str(Abort())==''`) — looks like a crash, not a cancel.

This isn't hypothetical — it's the exact Ctrl+C gesture the wizard's own intro banner invites ("Press Ctrl+C at any time to abort"), at exactly this field.

The codebase already has the fix for this pattern: `dango/cli/utils.py::safe_confirm()` (lines 92-133) wraps its own `click.prompt()` call in `try/except (click.Abort, EOFError)`. The PR's "Regression check" section claimed parity with the existing `raise click.Abort()` at line ~208, but that only fires on a deliberate "No" answer to a confirmation — not on the universal Ctrl+C gesture every other field in this function handles silently.

**Suggested minimal fix**, consistent with the sheet_selector branch's own "`if not answers: return None`  # User cancelled (Ctrl+C) - always abort" convention:
```python
elif param_name == "spreadsheet_url_or_id":
    ...
    default_str = str(default) if default is not None else None
    try:
        value = click.prompt(prompt, default=default_str, show_default=bool(default_str))
    except click.Abort:
        return None  # User cancelled (Ctrl+C) - always abort, same as every other field
    return value
```

### 2. New branch ignores the `required` parameter (currently inert, flagging for the record)

Every other branch in `_prompt_parameter()` uses `required` (appends "(optional)" to the message, returns `None` for empty non-required input, supports "Skip"). The new branch dispatches purely on `param_name` and never looks at `required`, so if `spreadsheet_url_or_id` (or this exact branch) were ever reused as an optional param, `click.prompt()` with no default re-prompts forever on empty input instead of returning `None`.

Currently inert: the registry's `spreadsheet_url_or_id` entry has no `required` key and is only ever collected via `required_params` (i.e., always `required=True`) — confirmed in `registry.py` lines 330-350. Not asking for a change here since it's out of the stated scope (this PR is deliberately scoped to this one field/this one call site), but worth a code comment or a `required` guard if this pattern is ever copied for another field.

**Recommendation:** fix #1 before merge (low-risk, same file, matches an existing established pattern). #2 can stay as a documented known-limitation given the current scope.

aaronteoh:
Fixed in 693f03e — caught \`click.Abort\` locally and return \`None\`, matching the \`if not answers: return None\` convention every other field uses for Ctrl+C.

Verified live via a real pty + real Ctrl+C (not just the unit test): before the fix, \`Spreadsheet ID or URL: ^C\` was immediately followed by a bare \`❌ Error: \` before the wizard's normal cancellation text; after the fix it goes straight to \`Source not added\` / \`Aborted!\`, identical to every other field. Full before/after detail is in the updated PR description.

Also flipped \`test_prompt_parameter_spreadsheet_url_ctrl_c_returns_none\` (previously \`..._raises_abort\`, which had been asserting the buggy behavior as correct) to assert the fixed behavior.

Finding #2 (ignores \`required\`) — left as-is per the earlier assessment: currently inert since the registry always sets \`required=True\` for this field, and out of this PR's stated scope.

PR #516, opened 2026-09-18
1.0.8-AH: Fix Metabase refresh restart timeout

aaronteoh:
## Summary
- `refresh_metabase_connection()` now tries `_wait_for_metabase_log_ready()`
  first (Q17's proven mechanism), falling back to the original
  `/api/health` poll only if the log check doesn't confirm readiness --
  mirrors AG's `setup_metabase()` fix (PR #514) applied to this sibling
  restart-readiness path.
- Budget changed from a fixed 20-iteration/1s-sleep guess (~20s, dating to
  2025-11-14, pre-1.0.8) to **60s**, derived from 4 live *restart*-path
  measurements (not AG's cold-first-boot numbers, which this task explicitly
  does not reuse unverified): wall-clock elapsed from the same `restart_time`
  capture point to Metabase's own `"Initialization COMPLETE ... (JVM uptime:
  Y.Ys)"` log line was **38.4s / 36.0s / 35.9s / 37.0s** across 4 real
  `docker restart` cycles against an already-warm, already-set-up container
  (real Docker container, real scratch project, non-default ports to avoid
  colliding with any other locally-running project). Max observed = 38.4s, a
  tight ~7% spread -- far tighter than AG's cold-boot 82.2-127.8s, consistent
  with a warm restart skipping image build/pull and benefiting from OS page
  cache. Chose **60s**: ~56% margin over the observed max, the same margin
  ratio AG's own 200s/127.8s local-mode value used for the cold-boot case,
  not a guessed round number and not a copy of AG's number.
- Login/session-ID capture/site-url-catchup logic unchanged. The old
  per-loop `logger.warning()` was replaced with an equivalent warning in
  the new structure, not dropped -- see "Review round 1" below.

## Verification
- `pytest -x`: 4747 passed, 32 skipped, 0 failed (full suite, run across
  this PR's lifecycle -- initial implementation, after removing temporary
  debug instrumentation used for the manual verification below, and again
  after the review-round-1 fixes)
- **Live timing measurement** (4 real warm-restart samples, container
  already healthy/warm before each restart via a fresh scratch project's
  first-run `setup_metabase_if_needed()`, non-default ports 8850/3050/8181):
  - sample1: 38.4s (JVM uptime 36.9s)
  - sample2: 36.0s (JVM uptime 33.9s)
  - sample3: 35.9s (JVM uptime 33.6s)
  - sample4: 37.0s (JVM uptime 34.7s)
- **Manual verification**: called the new `refresh_metabase_connection()`
  directly against the same real warm container with temporary debug
  instrumentation. Result: `log_ready=True health_ready=False`,
  `result=(True, None, '9ccb4afd-...')`, elapsed 38.1s -- confirms the log
  check fires and wins the short-circuit in practice (the `/api/health`
  fallback is never invoked), and that the **old 20s budget would have
  failed this exact call** (38.1s > 20s), which is the bug this PR fixes.
  Debug instrumentation removed before the final commit; the full-suite
  runs above bracket that removal.

## Regression check
- `_wait_for_metabase_log_ready()` and `wait_for_metabase_ready()` own
  definitions unchanged (grep-confirmed, not touched by this diff)
- All 8 call sites unchanged (`grep -n "refresh_metabase_connection(" dango/`
  confirms `dlt_runner.py`, `initial_sync.py`, `jobs.py`, `metabase_cmd.py`,
  `dashboard.py`, `transform.py`, `model.py`, `source.py` -- none inspect
  timing, only the returned tuple)
- Login/site-url-catchup logic inside the success branch unchanged (copied
  through verbatim)
- Two pre-existing tests (`test_uses_configured_url_from_creds_file_when_not_passed`,
  `test_explicit_metabase_url_wins_over_creds_file`) needed their
  `_wait_for_metabase_log_ready` mock flipped from `True` to `False` --
  they assert on the `/api/health` call's resolved URL, which is now only
  reached via the fallback path since a `True` log-ready short-circuits it
  entirely. This is a necessary consequence of the intentional
  restructuring, not a behavior change to what they test (URL resolution).
- **Accepted trade-off (not a code change):** worst-case blocking time on
  a *total* failure (Metabase genuinely never comes back) increased from
  ~20s (old fixed loop) to up to ~120s (60s log-check timeout + 60s
  `/api/health` fallback timeout, run sequentially). This is consistent
  with AG's already-merged, already-reviewed `setup_metabase()` precedent,
  which stacks to an even larger worst case (200s + 200s = 400s local,
  300s + 300s = 600s cloud) using the identical `_wait_for_metabase_log_ready(...)
  or wait_for_metabase_ready(...)` shape. Noting explicitly so a future
  timeout/scheduler audit isn't surprised by it.

## Review round 1
Independent review (posted as a PR review comment) found the removed
`logger.warning()` was not actually dead code: the scenario it covered
(Metabase's `/api/health` confirms readiness but the "Initialization
COMPLETE" log line never appeared) is still fully reachable -- it's exactly
what happens when the log check times out and the `/api/health` fallback
then succeeds. It had simply gone silent instead of becoming unreachable,
contrary to what this PR originally claimed.

Fixed:
- Restored an equivalent warning (reworded for the new structure, since it
  no longer implies *another caller's* login is at risk -- this function
  now does its own login unconditionally regardless of which check fired),
  gated on `health_ready and not log_ready` so it only fires for the
  fallback-succeeded case.
- Added a positive/negative-control test pair
  (`test_refresh_metabase_connection_falls_back_to_health_check` asserts
  the warning fires; `test_refresh_metabase_connection_succeeds_via_log_check`
  asserts it does not) proving the warning actually distinguishes which
  check fired rather than firing unconditionally.
- Restored the dropped rationale comment explaining why the log-line check
  is preferred over a raw `/api/health` poll (a still-shutting-down old
  process can answer 200 for a few seconds after `docker restart`'s SIGTERM
  before a real gap where nothing is listening).
- Added the accepted-trade-off note above (worst-case failure-path timing)
  to the Regression check section.

Claude-Session: https://claude.ai/code/session_01Nn6RWFt5T69QR6gvQ27igT


aaronteoh (commented):
Independent review (separate from the task-file/spec-conformance check already done). I traced the old vs. new control flow directly rather than re-checking against the spec. Three findings, none blocking correctness, but one I'd recommend fixing before merge:

## 1. Should-fix: the removed `logger.warning()` is not actually dead — its scenario is now silent

The Required-changes text (and my own PR description) claims the warning "warned about a case the new structure can no longer reach." I traced this and it's not accurate.

Old code: on `/api/health` 200, it *also* waited (blocking, up to 90s default) for `_wait_for_metabase_log_ready()`. If that timed out, it logged a warning ("restart succeeded and /api/health responded, but ... log line never appeared") and still returned success.

New code: `log_ready` is tried first (60s). If it fails, `/api/health` fallback runs (60s). If that succeeds, `health_ready=True` and we return success via the exact same substantive scenario — health confirms, log line never appeared — just reached via the fallback branch instead of the old nested check. **No log output occurs for this case at all now.** Given this project's own history needing exactly this kind of signal to diagnose the Q11/Q14/Q16/Q17 readiness-race saga, losing it silently seems like a real (if minor) observability regression, not a no-op removal.

Suggested fix: an `elif health_ready:` branch (or equivalent) logging that the fallback fired, distinguishing it from the normal log-check-succeeded path.

## 2. Worth confirming acceptable: worst-case failure-path blocking time went from ~20s to up to ~120s

Previously, a total failure (Metabase never comes back) returned in ~20s (`max_attempts=20 * 1s`). Now it's up to 60s (log check) + 60s (health fallback) = 120s before returning failure. This mirrors AG's `setup_metabase()` shape exactly (which has an even larger worst case, 200s+200s local / 300s+300s cloud), so it's a consistent, precedented pattern — but neither the Regression risk section nor the PR description called out this specific angle (they covered timeout-value accuracy and the log-only-no-health-check case, not the failure-path duration itself). Worth a one-line note for whoever next audits scheduler/CLI timeout budgets, since a scheduled job or web request blocking 2 minutes on a hard failure is a different operational shape than 20s.

## 3. Nit: lost rationale comment

The removed inline comment explained *why* the log-check is preferred over `/api/health` (a shut-down old process can keep answering 200 for a few seconds post-SIGTERM). The new top-of-block comment only covers timing/margin derivation, not that rationale. Minor, but it's exactly the context a future maintainer would need to notice finding #1/#4 below on their own.

## 4. Informational only, not a bug: fallback path has no cross-check against the log signal

Once `log_ready` fails, `health_ready` is set on a single raw `/api/health` 200 with no further log confirmation — in principle reopening the zombie-process race the log check exists to close. In practice this window only opens 60+s after the restart was issued, well past any realistic post-SIGTERM zombie-response period, and it's the exact same shape as AG's `_wait_for_metabase_log_ready(...) or wait_for_metabase_ready(...)`. Not asking for a change — just noting the Regression risk section only pre-cleared the "log_ready alone, no health check" side of this, not this side.

Everything else checked out: `_apply_metabase_site_url_catchup` (lines 416-504) traced directly to confirm it makes no `session.get()` calls in the new tests' bare-`tmp_path` path, so `mock_get.assert_not_called()` in `test_refresh_metabase_connection_succeeds_via_log_check` is a sound assertion, not a false negative. Login/site-url-catchup/return logic is copied verbatim. Call-site grep and margin-ratio math both check out.

